PR #60, opened 2021-05-12
Create smoketest script

HonkingGoose:
## Changes

- Creates a "smoketest" script

## Context

The idea behind this script is that you use it on your local workstation to check if you're good to go to create a pull request or push to `main`.

- Option 1: script only checks for linting errors.
- Option 2: script fixes linting errors where possible, so you don't need to run ESLint and Prettier in "fixup" mode yourself.

The current PR uses option 1.

Let me know if you like option 2 better, and I'll change the PR. 😉

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "dev": "next",
     "build": "next build",
     "serve": "next start",
+    "smoketest": "yarn type-check && yarn eslint && yarn prettier:check && yarn build",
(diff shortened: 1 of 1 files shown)


Belco90 (approved):
I'm happy with option 1, I like this kind of scripts to check things but let me fix them (and we even have other commands to fix autofixable issues).

PR #62, opened 2021-05-12
improve call to action text on about page

HonkingGoose:
## Changes

- Improve call-to-action text

## Context

The new text is clearer on the desired action "Try me now!", and also makes it clearer where clicking the link next to the button will take the reader.


Belco90 (approved):
Nice, thanks!

PR #87, opened 2021-05-20
Improve CircularProgress comnponent

HonkingGoose:
## Changes

- Add explicit `trackColor`
- Reduce size of `CircularProgress `component
- Darken indicator color

## Context

Some small improvements to the CircularProgress component.
This is the ring with indicator that you see while Octoclairvoyant is processing the search for the repository name.


HonkingGoose:
Force pushed to fix up the bad commit message. I'll stop moving around now, so that you can review it. 😄

Belco90 (approved):
Looking great, thanks for the improvement!

PR #88, opened 2021-05-20
Fix smoketest script

HonkingGoose:
## Changes

- Fix `smoketest` script

## Context

Turns out I used the wrong script names in PR #60... 🙈

Please check on your end if this does what you want it to, and that it works properly. 😉


Belco90:
Tested locally, it works like a charm!

Belco90 (approved):

PR #94, opened 2021-05-23
Create bug and feature issue templates

HonkingGoose:
## Changes

- Create a bug report template
- Create a feature suggestion template
- Prevent contributors/bug reporters from opening a blank issue with `config.yml` file

## Context

Having a template helps bug reporters or contributors to give the correct information.
Both issue templates will use an appropriate label for their respective issues.

I've also made it so that contributors cannot open a empty issue, so they must use either the bug report template, or feature suggestion template.

Feel free to suggest a different wording. 😉


Belco90 (approved):
Good idea! Looks good to me 👌

PR #106, opened 2021-05-29
Version tag badge links to GitHub release

### LINKED ISSUE #89: Version tag badge links to GitHub release
## Problem

Right now the change log does not provide an easy way to jump to the GitHub tagged release.
We show a badge with the GitHub tagged version, but clicking on this badge does nothing.

The Octoclairvoyant back-end knows about the tagged versions (because it's filing out the badges with the tag version), but we're not surfacing this to the end-user in an easy way.

## New feature

The version badges are links that the user can click on.
When the user click on the version badge, a new browser window opens with a link to the GitHub release for that version.

## Proposed behavior

1. Go to the octoclairvoyant webpage.
2. Enter search string `renovate`.
3. Select the `renovatebot/renovate` repository from the drop-down.
4. Select version, from `24.0.0` to `25.0.0`.
5. Click on the version badge for `25.0.0`.
6. New browser window opens to: https://github.com/renovatebot/renovate/releases/tag/25.0.0

## Further context

I don't know how hard/easy this is to implement? Feel free to close this if this is too much work for you. 😉

Belco90:
## Changes

- Add a link to the release version badge

## Context

Closes #89

### DIFF
--- a/src/components/ProcessedReleaseChangeDescription.tsx
+++ b/src/components/ProcessedReleaseChangeDescription.tsx
@@ -90,7 +90,15 @@ const ProcessedReleaseChangeDescription = ({
         <>
           <Tag size="md" mb={2} rounded="full" colorScheme="primary">
             <Box as={GoTag} h={4} w={4} mr={1} />
-            <TagLabel>{getReleaseVersion(processedReleaseChange)}</TagLabel>
+            <TagLabel>
+              <Link
+                href={processedReleaseChange.html_url}
+                isExternal
+                color="primary.800"
+              >
+                {getReleaseVersion(processedReleaseChange)}
+              </Link>
+            </TagLabel>
           </Tag>
           <Box ml={4}>{processedDescription}</Box>
         </>
--- a/src/components/ProcessedReleaseChangeDescription.tsx
+++ b/src/components/ProcessedReleaseChangeDescription.tsx
@@ -12,7 +12,6 @@ import {
   Text,
 } from '@chakra-ui/react'
 import { FiExternalLink } from 'react-icons/fi'
-import { GoTag } from 'react-icons/go'
 import { ComponentProps } from 'rehype-react'
 
 import BlockQuote from '~/components/BlockQuote'
@@ -98,7 +97,6 @@ const ProcessedReleaseChangeDescription = ({
               _hover={{ bgColor: 'gray.300' }}
               _active={{ bgColor: 'gray.400', color: 'gray.900' }}
             >
-              <Box as={GoTag} h={4} w={4} mr={1} />
               <TagLabel>{getReleaseVersion(processedReleaseChange)}</TagLabel>
             </Tag>
           </Link>
--- a/src/components/ProcessedReleaseChangeDescription.tsx
+++ b/src/components/ProcessedReleaseChangeDescription.tsx
@@ -90,12 +90,13 @@ const ProcessedReleaseChangeDescription = ({
         <>
           <Link href={processedReleaseChange.html_url} isExternal>
             <Tag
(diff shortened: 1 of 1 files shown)


Belco90:
@HonkingGoose let me know what you think!

src/components/ProcessedReleaseChangeDescription.tsx
```diff
@@ -90,7 +90,15 @@ const ProcessedReleaseChangeDescription = ({
         <>
           <Tag size="md" mb={2} rounded="full" colorScheme="primary">
             <Box as={GoTag} h={4} w={4} mr={1} />
-            <TagLabel>{getReleaseVersion(processedReleaseChange)}</TagLabel>
+            <TagLabel>
```
HonkingGoose, line 93: Cool! I like this! 😄 

I have 2 suggestions:

- make the entire tag area a link/clickable.
- Add `:focus` and `:hover` styles

Only the link text is clickable, normally in an interface you can click on the whole button. Can you make it so that the whole tag area is clickable?

I'd also like `:hover` and `:focus` states for the tag area, so that when you hover over the badge, there's a slight color change. Would be nice to have a `focus-ring` on the outline of the badge when it's selected via keyboard.

HonkingGoose (changes requested):

src/components/ProcessedReleaseChangeDescription.tsx
HonkingGoose: Oh, and the background of the tag could maybe change as well once you actually click on it with `:active`. 😄 

Chakra UI has the wanted props! 😄 

- `_hover`
- `_focus`
- `_active`

Source: https://chakra-ui.com/docs/features/style-props#pseudo

Belco90: Making the entire tag clickable seems much better indeed. I'll play around with it and push something during the week.

HonkingGoose: > Making the entire tag clickable seems much better indeed. I'll play around with it and push something during the week.

Glad you agree. I look forward to the improvements. 😄

Belco90: Updated! Let me know what you think.

HonkingGoose:
I've pushed some changes, let me know what you think of them. 😉 

### De-emphasized version badges

De-emphasized version badges by using a neutral gray color, I think too much color is distracting from the text of the changelog. I view the version badges (and the commit hash/issue) link as tertiary content. You should be able to find it when needed, but otherwise it should not be drawing attention to itself.

### Removed tag icon

Remove tag icon, as I think it looks cleaner without the icon, I think the context of the number is also clear, as users will have selected a version range in the repository select component already.
If you do want to keep the icon, I'll push a new commit which de-emphasizes the icon, so it doesn't draw too much attention.

Belco90:
@HonkingGoose I'm happy with the changes you did. Agree on the de-emphasized tags, probably taking more attention than they should indeed.

HonkingGoose:
> @HonkingGoose I'm happy with the changes you did. Agree on the de-emphasized tags, probably taking more attention than they should indeed.

By the way, I don't propose we make a full gray-scale site. The final design will use the Octoclairvoyant brand colors in suitable areas! 😉

HonkingGoose (approved):

Belco90:
> By the way, I don't propose we make a full gray-scale site. The final design will use the Octoclairvoyant brand colors in suitable areas! 😉

Sure! I just put the brand colors here and there when building the webapp without checking if made sense or not to be honest, so happy to improve that.

PR #107, opened 2021-05-29
Update Prettier to v2.3.0

HonkingGoose:
## Changes

- Use Prettier 2.3.0
- Format files with the newer version of Prettier

## Context

You closed an earlier Renovate bot upgrade to Prettier `2.3.0`, so maybe you really don't want this update. But I figured I'd try again. 😄 

If you don't want to upgrade, please close the PR, and also explain why you don't want to upgrade now, so that I know the reasoning. 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "jest": "27.0.1",
     "lint-staged": "10.5.4",
     "pinst": "2.1.6",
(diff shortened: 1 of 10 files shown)


Belco90 (approved):
I closed that renovate PR because we needed also to format all files, which you did already. So this is good to go!

PR #132, opened 2021-06-06
Improve about page

HonkingGoose:
## Changes

- Different list icons
- Different button color
- Remove background gradient, replace with `gray.50` background
- Use padding to get a bit of whitespace on either side of the content
- De-emphasize icons by using a lighter color
- Make headings a bit nicer by tightening the letter spacing
- Remove link to features list, as we can now always see the feature list in the window

## Context

I wanted to see if I could improve the About page we have now, and make it look more like the Figma mockups.
I know the about page will get replaced with a nice homepage later on, but we could still make what we have now look a bit better. 😉

### DIFF
--- a/src/pages/about.tsx
+++ b/src/pages/about.tsx
@@ -13,40 +13,52 @@ import {
 import { NextSeo } from 'next-seo'
 import Image from 'next/image'
 import NextLink from 'next/link'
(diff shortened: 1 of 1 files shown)


Belco90 (approved):
Looks good to me!

PR #138, opened 2021-06-08
Add active state style to CTA button

HonkingGoose:
## Changes

- Add active state style to call to action (CTA) button

## Context

Forgot to override the default active state color of the button (it used to be a green color, now it's purple).

Fixes up a missed part of PR #132.


Belco90 (approved):
👌

PR #176, opened 2021-06-25
Automerge minor and patch Jest updates

HonkingGoose:
## Changes

- Automerge minor and patch level Jest updates

## Context

I don't know if you want to manually verify the Jest tests? (I don't think we actually **run** any Jest tests at all right now?)

We should still manually check the changelogs for major Jest updates, as those are likely to have major new features/breaking changes.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -18,6 +18,13 @@
       "matchDepTypes": ["devDependencies"],
       "matchPackagePatterns": ["eslint", "prettier"],
       "automerge": true
+    },
+    {
+      "description": "Automerge minor and patch Jest updates",
+      "matchUpdateTypes": ["minor", "patch"],
+      "matchDepTypes": ["devDependencies"],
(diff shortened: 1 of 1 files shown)


Belco90 (approved):
We could even uninstall jest for now until is used for real.

HonkingGoose:
> We could even uninstall jest for now until is used for real.

I tried getting Jest setup to test our components, but this is too complex for me to do.
Also it seems there is no documentation on working with Next.js + Jest specifically, so I don't know what the best practice would be.

I did find Cypress.io, that seems a really nice test app. Maybe you can take a look at that and see if that might fit better than using Jest?

https://www.cypress.io/

HonkingGoose:
I opened an issue to discuss creating some actual tests (#177).

Maybe it's best to just remove the Jest dev dependency until we're actually using it. 😄

PR #186, opened 2021-06-28
Initial setup for new color scheme

### LINKED ISSUE #185: Prepare to use new color scheme in our code

The end-goal is to use the colors from the approved design.
Before we can use any of these new colors, they'll need to be made available to the backend.

Central source of truth for our color use in the app.

I'm already working on this. 😉

HonkingGoose:
**This PR assumes that you want to use the new color scheme I picked.... 😉**

Marking this as draft until we're sure what colors we want to use.

I think this is what we would need to do to get the colors ready for use?

---

## Changes

- Create constants for our new colors `coolGray`, `fuchsia` and `sky`
- Create constants for `primaryText`, `secondaryText` and `tertiaryText` and their darkmode equivalents
- Add comments explaining how I intend us to use these colors

## Context

Fixes #185.


Belco90:
Since this is a draft I haven't paid attention to it. Let me know if you are expecting some feedback from it.

HonkingGoose:
The idea is to get the colors ready for use, but I don't know if this is the right place to put the new colors?

Belco90:
Yes, it is the right place, although you should replace existing colors with those new ones added. [Here](https://chakra-ui.com/docs/theming/theme#colors) you can find existing colors in Chakra.

So if we want `tailwind_coolGray` to be the default gray, you need to assign it to `gray` key. I'll add a suggestion to the code for clarification.

src/customTheme.ts
```diff
@@ -39,11 +45,80 @@ const blueColor: ColorHues = {
   900: '#00131b',
 }
 
+// ***** New colorscheme *****
+
+// The coolGray palette contains all the shades of gray we'll use in the app.
+// Example uses: body text, headings, changelog text, page background.
+// https://tailwindcss.com/docs/customizing-colors#color-palette-reference
+
+const tailwind_coolGray: ColorHues = {
```
Belco90, line 54: [ESLint `camelcase` rule](https://eslint.org/docs/rules/camelcase#require-camelcase-camelcase) is not enabled but we should have all the constants purely in camelCase 😅 

FYI: the rule is not enabled since we receive keys in snake_case from GitHub API and the rule would complain there. Something I could solve by mapping the keys received, but I didn't have time to do that.

src/customTheme.ts
```diff
@@ -39,11 +45,80 @@ const blueColor: ColorHues = {
...
     blue: blueColor,
     primary: primaryColor,
     secondary: blueColor,
+    coolGray: tailwind_coolGray,
+    fuchsia: tailwind_Fuchsia,
+    sky: tailwind_Sky,
+    primaryText,
+    secondaryText,
+    tertiaryText,
   },
```
Belco90, line 122: Since we are interested in overwriting default colors with the new ones you defined, here you need to reassign them instead of assign them to new keys:

```suggestion
  colors: {
    blue: tailwind_Sky,
    primary: tailwind_Fuchsia,
    secondary: tailwind_Sky,
    gray: tailwind_coolGray,
    purple: tailwind_Fuchsia, // assuming we want to replace default `purple`
  },
```

Not really sure how to change the text colors, I need to investigate those.

src/customTheme.ts
HonkingGoose: So from `tailwind_coolGray` (snakecase) ->`tailwindCSSCoolGray` (camelcase) then?

src/customTheme.ts
HonkingGoose: I don't know if this is a good or bad idea. It's certainly "short-term easiest", as we're basically replacing all colors at once.... 😄 It won't look any good until I go over each component, and set the proper hue. Like for example: change `gray.500` to `gray.300` to account for the different hues.

I worry that clobbering over the Chakra UI defaults in this way will lead to problems later on. My initial plan was to change "old color" with "appropriate new color" by hand, by replacing the old variables with the new tailwindCSS variables.

What do you think?

src/customTheme.ts
Belco90: That's right, or even `tailwindCoolGray` for simplicity!

HonkingGoose: > tailwindCoolGray

That's a good name!

src/customTheme.ts
Belco90: That's ok! The good thing here is that we can add extra colora or replace default ones. I thought we wanted the latter. But if you only want to add new colors for now, then the current code is fine.

Even if we are not replacing default ones now, the goal would be to do so at some point right? Otherwise having tons of color tokens would add unnecessary complexity, specially if we don't use the main tokens.

HonkingGoose: The desired end result is that we only use these 3 TailwindCSS color palettes once we're fully done:

- `coolGray`
- `fuchsia`
- `sky`

I don't want any other colors to be used. It **will** look bad if a component mixes the new Tailwind colors and their own "old" Chakra colors.

There are two ways to get to that end-result, I think.

Options:

1. Clobber over the old color all at once like in your suggestion
2. Gradually introduce the new colors by manually replacing the reference from `oldColor` -> `newTailwindColor`

Like I said, I'm not sure which way is best.

Either way, it will need manual work, because the color schemes are too different from each other. 😄

src/customTheme.ts
```diff
@@ -39,11 +45,80 @@ const blueColor: ColorHues = {
...
+  900: '#0C4A6E',
+}
+
+// These are our normal text colors.
+// Do NOT use when you have text that goes on a colored background.
+// Text on colored backgrounds should use a suitable hue from the background color scheme.
+
+const primaryText = '#111827'
+const secondaryText = '#4B5563'
+const tertiaryText = '#6B7280'
```
HonkingGoose, line 107: I just realized I need to figure out 3 good shades of gray for use on our darkmode as well.
That way all possible text colors are nicely systematized. 😄 

```js
// 3 text colors for use with light mode

const primaryTextLightmode
const secondaryTextLightmode
const tertiaryTextLightmode

// 3 text colors for use with dark mode

const primaryTextDarkmode
const secondaryTextDarkmode
const tertiaryTextDarkmode
```

HonkingGoose: The text colors will al be a shade of coolGray (I think). Is there a way to just link to the correct `coolGray` shade? I'm looking to do something like this:

```typescript
const primaryText = tailwindCoolgray.900
```

Belco90: You definitely can assign it like that, just changing the syntax a bit: `const primaryText = coolGray['900']`

I'm not sure how text colors are assigned yet, so I need some time to investigate that. Anyway, just focus on finding the proper colors so we can use them manually in some components, and then I can set it as the corresponding default token.

src/customTheme.ts
Belco90: I'm happy to gradually introduce the new colors! Just wanted to make sure our goal would be to replace the default ones with those.

> The desired end result is that we only use these 3 TailwindCSS color palettes once we're fully done:
> 
>     * `coolGray`
> 
>     * `fuchsia`
> 
>     * `sky`

This is ok. However, I'd like to do this over some default tokens, that's what I wanted to express. So in Chakra components we use:
- `gray` token --> `coolGray` color
- `primary`/`purple` token --> `fuchsia` color
- `secondary`/`blue` token --> `sky` color

Anyway, I guess our plan here is:
- include new colors with new tokens (you already did this)
- start using the new colors bit by bit in the components over the codebase
- override default tokens with new tokens

src/customTheme.ts
```diff
@@ -26,6 +30,8 @@ const primaryColor: ColorHues = {
   900: '#180319',
 }
 
+// The old blueColor palette will be replaced with a suitable hue out of the Sky TailwindCSS color palette.
+
 const blueColor: ColorHues = {
   50: '#def8ff',
```
Belco90, line 36: I just realized the current colors are defined incorrectly: the keys should be strings (`'50'`, `'100'`, etc) instead of numbers (`50`, `100`, etc). I'll create a small PR later to fix this.

HonkingGoose:
> I just realized the current colors are defined incorrectly: the keys should be strings (`'50'`, `'100'`, etc) instead of numbers (`50`, `100`, etc). I'll create a small PR later to fix this.

@Belco90 Can you confirm if my new commit fixes this problem?

HonkingGoose:
Marked this ready to go a bit too soon, we still need to figure out how we want to replace the colors.

I'll go make a separate PR with the contents of this PR, where I change the colors all in one go, so we can review how that looks. 😉

Belco90:
> > I just realized the current colors are defined incorrectly: the keys should be strings (`'50'`, `'100'`, etc) instead of numbers (`50`, `100`, etc). I'll create a small PR later to fix this.
> 
> @Belco90 Can you confirm if my new commit fixes this problem?

Partly: the declaration is fine but the type must be updated. Instead of using our own `ColorHues` type we should import it from `@chakra-ui/react`. That's what I wanted to do in a separate PR since I'm not sure how experienced you are with TS.

HonkingGoose:
> Partly: the declaration is fine but the type must be updated. Instead of using our own `ColorHues` type we should import it from `@chakra-ui/react`. That's what I wanted to do in a separate PR since I'm not sure how experienced you are with TS.

Opened a new issue to track this problem, I agree we should do the proper thing, but for now let's focus on getting the new colors ready for use. 😉

HonkingGoose:
Hi @Belco90 I've updated this PR so that we have the _color hue typing improvements_ from #239 here as well.

Can you review this PR?

HonkingGoose:
> Anyway, I guess our plan here is:
> 
>     * include new colors with new tokens (you already did this)
> 
>     * start using the new colors bit by bit in the components over the codebase
> 
>     * override default tokens with new tokens

I think I like this plan best, clobbering over all colors at once does work (#218), but it's a bit of a hack job. I'd rather go through the colors one by one, and make sure the colors are the same as on my Figma design. 😄 

Do you have any thoughts/feedback on this plan?

Belco90:
> Hi @Belco90 I've updated this PR so that we have the _color hue typing improvements_ from #239 here as well.
> 
> Can you review this PR?

Sure. I'll try to take a look later.

Belco90:
> > Anyway, I guess our plan here is:
> > ```
> > * include new colors with new tokens (you already did this)
> > 
> > * start using the new colors bit by bit in the components over the codebase
> > 
> > * override default tokens with new tokens
> > ```
> 
> I think I like this plan best, clobbering over all colors at once does work (#218), but it's a bit of a hack job. I'd rather go through the colors one by one, and make sure the colors are the same as on my Figma design. 😄
> 
> Do you have any thoughts/feedback on this plan?

I like this plan! I guess you'll start working on step 2 after this MR gets ready, and then I can take care of the last step when all colors are used properly in the project. I've been reading Chakra documentation and I think I know how to do it now.

Belco90 (dismissed):
This looks good to me!

HonkingGoose (commented):
I'm happy with this. I think we can merge after your final review. 😉

Belco90 (approved):

HonkingGoose:
Hooray! Congratulations to ourselves for getting this merged! 😄 🚀 

I've made two new issues so that we can track the rest of our migration plan (#243 and #244).

PR #195, opened 2021-06-30
Cache yarn packages

HonkingGoose:
## Changes

- Cache yarn packages on our GitHub Action runners

## Context

It's easy to add caching now, as it's built-in to the action.
Used to be you had to use a separate caching action. 😄 

Important note from [actions/setup-node v2.2.0 release notes](https://github.com/actions/setup-node/releases/tag/v2.2.0):

> Yarn caching handles both yarn versions: 1 or 2.
>
>    At the moment, only `lock` files in the project root are supported.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,7 @@ jobs:
         uses: actions/setup-node@v2.2.0
         with:
           node-version: ${{ matrix.node-version }}
+          cache: "yarn"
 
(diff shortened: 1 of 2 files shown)


HonkingGoose:
I know our action output says: "yarn cache not found" but that's because this is the first time the action runs with the cache option enabled. It should find the cache the next times. 😉

Belco90 (approved):
Thanks!!

PR #197, opened 2021-06-30
Inter font, weights: 400, 700, 900 (regular/bold/black)

### LINKED ISSUE #184: Install/import new font Inter into our code

The design uses the Inter font, so our implementation should use Inter as well.
We should only ever use the Inter font, I don't want to use other fonts, if at all possible.

~We can use `@fontsource/inter` to install the font~, or we can use Google Fonts.

We should set things up so that the only font that's ever in use on our app is the Inter font.
I don't want any Chakra-UI prebuild-components to use some other font. 😉 

Makes the text consistent throughout the app, and uses a font designed for reading on a digital screen.

Ofcourse!

I'm not sure what I would need to change though.

I found references to open sans in these places:

https://github.com/Belco90/octoclairvoyant/blob/41dfc079044f1a88f0f6d7683cb8a814e33f66d2/src/pages/_app.tsx#L10-L11

https://github.com/Belco90/octoclairvoyant/blob/41dfc079044f1a88f0f6d7683cb8a814e33f66d2/src/customTheme.ts#L42-L56

HonkingGoose:
## Changes

- Load Inter font weights 400, 700 and 900 (regular/bold/black)
- Adjust some text in the app to look a bit better by changing `fontWeight` and `letterSpacing`

## Context

It doesn't look very good yet, but we need to start somewhere. 😄

Fixes #184.


src/components/Header.tsx
```diff
@@ -118,7 +118,11 @@ const Header = (props: BoxProps) => {
                 placeholder="blur"
               />
             </Box>
-            <Heading fontSize={{ base: 'md', md: 'xl', lg: '4xl' }}>
+            <Heading
+              letterSpacing="tight"
```
HonkingGoose, line 122: The _Inter_ font isn't designed to do work as a "display font" that used with big font sizes. I'm tightening the letter-spacing to condense the font, so it looks a bit better.

src/customTheme.ts
```diff
@@ -46,7 +46,9 @@ const customTheme = extendTheme({
     secondary: blueColor,
   },
   fonts: {
-    heading: '"Open Sans", sans-serif;',
+    heading: '"Inter", sans-serif;',
+    body: '"Inter", sans-serif;',
+    mono: '"Inter", sans-serif;',
```
HonkingGoose, line 51: This should get us using the _Inter_ for most things...

You said this wasn't a complete replacement of the Chakra UI default font? So we'll probably need to do more work later. But for now it's a good start, I think?

src/pages/_document.tsx
```diff
@@ -7,7 +7,7 @@ class MyDocument extends Document {
         <Head>
           <link
             rel="stylesheet"
-            href="https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700&display=swap"
+            href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700;900&display=swap"
```
HonkingGoose, line 10: Loading in 400, 700 and 900 for now.

400 (regular) for normal body text, 700 (bold) for emphasis, and 900 (black) for our hero/big text.

HonkingGoose (commented):
Commenting with my ideas/thoughts on this. 😄

src/customTheme.ts
Belco90: This covers all font tokens from Chakra! Unless we want to add more tokens for fonts (I don't think so right?) this is all we need to do.

Belco90 (approved):
It looks great, thanks for your PR!

src/customTheme.ts
Belco90, line 51: Going through this again, I think we should avoid changing the mono one. This one is used in code blocks, which look weird with Inter font. An example here: https://octoclairvoyant-webapp-git-load-inter-font-octoclairvoyant-team.vercel.app/?repo=testing-library%2Fdom-testing-library&from=v6.11.0&to=latest

What do you think?

Belco90 (changes requested):

src/customTheme.ts
HonkingGoose:
```suggestion
```

I agree, the Inter font doesn't look nice on the code blocks! The Inter font does not come in a monospaced variant yet, but there's a popular issue for it on their GitHub issue tracker. So maybe my favorite font will be suitable for code-blocks later on!

The current production site uses the system operating font stack, which on macOS is Menlo Regular, and on Windows Segoe UI. Those fonts are not a monospaced font by the way.

I think the proper fix is:

- Revert using Inter for monospaced font with my suggestion
- Review this PR, and merge if happy
- I'll go search for a nice monospaced font that matches well with the Inter font, and make a PR to load in that font

Belco90: Happy to do that!

HonkingGoose:
@Belco90 can you take a look at this and tell me what you think?

https://www.smashingmagazine.com/2019/06/optimizing-google-fonts-performance/

Search for `DNS Prefetching` and `Preconnect`.

Maybe using this can be faster? I noticed when I grabbed the Inter font from Google Fonts, that the `<link>` tag came filled with those properties.

But maybe Next.js 11 does DNS Prefetching and Preconnect on the fonts automatically, so we don't need to put those things in our `<link>` tags?

Belco90:
> @Belco90 can you take a look at this and tell me what you think?
> 
> https://www.smashingmagazine.com/2019/06/optimizing-google-fonts-performance/
> 
> Search for `DNS Prefetching` and `Preconnect`.
> 
> Maybe using this can be faster? I noticed when I grabbed the Inter font from Google Fonts, that the `<link>` tag came filled with those properties.
> 
> But maybe Next.js 11 does DNS Prefetching and Preconnect on the fonts automatically, so we don't need to put those things in our `<link>` tags?

I assumed Next.js 11 was adding this automatically but it seems it's not! I can see the preconnect one being added in production at the moment, but their new ESLint rules should complain if you don't add it manually. I think they are working on handling this automatically, but better to include it manually for now.

Belco90:
I'm confused now. Those 2 resource hints shouldn't be necessary since [Next.js Font Optimization](https://nextjs.org/docs/basic-features/font-optimization) "will automatically inline font CSS at build time, eliminating an extra round trip to fetch font declarations". So we shouldn't need to worry about those prefetching and preconnect hints, but I'm not sure if the font is getting inlined as expected.

HonkingGoose:
I went searching trough the Next.js repo issues and discussions, maybe that will help understand the current state of things?

- https://github.com/vercel/next.js/discussions/24645 
- https://github.com/vercel/next.js/issues/24781

HonkingGoose:
How about we merge what we have now, and make a new issue to figure out how the optimization works? That way we can at least make progress on using better fonts. 😄

Belco90 (approved):

Belco90:
Sure! I didn't realize you already reverted the mono one.

Belco90 (approved):

HonkingGoose:
> Sure! I didn't realize you already reverted the mono one.

It's easy to miss in all the comment soup. 😉 

I'll go make a separate issue to figure out the font optimizations.

PR #199, opened 2021-06-30
Eslint plugin import improvements

Belco90:
## Changes

Set up `eslint-plugin-import` correctly so:
- we enable recommended rules preset
- we sort aliases properly


HonkingGoose (approved):
I think it looks OK, from what I can tell by reading the `eslint-plugin-import` readme.

PR #204, opened 2021-07-02
Use Roboto monospaced font

### LINKED ISSUE #200: Use nice monospaced font from google fonts

While reviewing PR #197, we noticed that the Inter font looks pretty bad when used for our code blocks.

We need to use a font that's designed to display code. Preferably one that matches well with the looks of Inter.

Get in control of how the app looks, because now everybody is using the same font.
Right now it seems Chakra UI defaults to the system font stack on the code blocks, but this will be different on macOS/Window/Linux.

Yup, I'll go find a nice font for us to use, and make a PR to load it in. 😄

## Fonts to try:

- [x] Roboto mono
- [x] Source Code Pro

## Rejected:

- [x] Ubuntu mono
- [x] Roboto (normal font)
- [x] IBM Plex mono (too busy to render code)

HonkingGoose:
## Changes

- Load in Roboto monospaced font (400, 700, and Italic variants)

## Context

Fixes #200.


HonkingGoose:
Hi @Belco90,

I've selected 2 fonts for us to choose from, we can use **Roboto**, or **IBM Plex Mono**.
I think the **Roboto** font looks best, so I'm showing you that one first. 😉

Belco90:
Not really convinced with Roboto to be honest. Here you have a comparison against the current version:
  
Current mono font vs New mono font

The current one is easier to ready in terms of code, which is the main usage of the `mono` font family in the webapp. Roboto seems too small not really code-ish.

HonkingGoose:
> The current one is easier to ready in terms of code, which is the main usage of the `mono` font family in the webapp. Roboto seems too small not really code-ish.

I think Chraka UI decreases the fontsize for monspaced text somehow... and it makes everything look too tiny/bad.

It looks a lot better when you use the same font-size for all normal text. Try using the browser dev-tools to increase the font to the same size.

The designer of the Inter font recommends Roboto or IBM plex as good matches with Inter. So that's why I'm trying those.

I'll push up the IBM plex branch now, so you can see how you like that one. 😉

Belco90:
> The designer of the Inter font recommends Roboto or IBM plex as good matches with Inter. So that's why I'm trying those fonts first.

I totally get this point, but we are interested in mono fonts with nice punctuation marks for making it easier to read the code, so I'm not sure if those suggestions are what we are looking for even if they are good matches with Inter.

HonkingGoose:
Turns out I was using the normal Sans-serif Roboto font, instead of the intended monospace font.

Maybe you like this one better! 😄

Belco90 (approved):
Much better now, I like this one! Are you happy with it?

HonkingGoose:
I like it very much as well. Though I do want to tinker with the font size a bit later on! 😄 

Let's get this one in the code, so we can improve on it later.

PR #215, opened 2021-07-04
Use IBM Plex Mono font

### LINKED ISSUE #200: Use nice monospaced font from google fonts

HonkingGoose:
## Changes

- Load IBM Plex mono font (400, 700 and Italic variants)

## Context

Fixes #200.

Let's see how you like IBM Plex then. 😉


Belco90:
This one seems better in terms of font size and code readability, but still seems _too fancy_ for rendering code.

HonkingGoose:
I think IBM Plex Mono is too busy also... So let's remove this one from the list of possible fonts then. 😉

PR #218, opened 2021-07-05
Throwaway PR do not merge, change all colors at once

HonkingGoose:
## Changes

- Let's try replacing all old stuff with new stuff in one go

## Context

Checked this with the realistic query string, and it looks "OK".

If we want to go this way, I'll need to tune some component colors before we land it, because right now some shades of fuchsia and sky are way too bright. 😄


HonkingGoose:
I think I'd rather change the old colors to the new colors by hand. It's going to need adjustments anyways. 😄

PR #221, opened 2021-07-05
Create CONTRIBUTING.md

### LINKED ISSUE #220: Pick realistic comparator input to check our code with

I'd like to have a standard comparator query-URL that we will then use whenever we're comparing a new piece of code to some old piece of code. This way it's easy to see what works in the design, and what doesn't work. Because we always use the same string, its easier to see what changes between versions.

The best thing to have would be a query-string that has realistic output, and uses as much of the different components as possible (codeblocks, headings, changelog text, etc...).

Makes it easier for me to see what I need to change in the code to refine the design. 😄 

Have the link in a central place, so I don't have to go hunting in our issues for a suitable query string.

I can't because I don't have a good idea of what the realistic testing input should be.

We should put this URL somewhere in a development document. 😄

HonkingGoose:
## Changes

- Create a basic `CONTRIBUTING.md` file with instructions on how to get started

## Context

We should have a document that explains how to get started with helping us.
This is beneficial for us too, as we can put things we need to remember while developing in here. 😉 

Fixes #220.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributing
+
+Thank you for wanting to help us!
+Read this document for information on how to get started.
+
+## General overview of project
+
+We use [Next.js](https://nextjs.org/) to generate the static website.
+All our website styling is done with the [Chakra UI](https://chakra-ui.com/) framework.
+
+## What you'll need to know to be able to help
+
+You'll need a basic understanding of HTML/CSS to be able to help with the code.
+It's even better if you know a bit of React and the basics of TypeScript.
+
+## Software requirements
+
+- Git
+- Latest version of Node.js 14
+- Yarn v1 package manager
+
+This guide assumes you'll run the code on your workstation directly.
+
+### Where to get Git
+
+The [git-scm.com](https://git-scm.com/) site has links to download Git for your operating system.
+
+### Where to get Node.js
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+1. Create a feature branch
+1. Start the development server with `yarn start`
+1. Make improvements
+1. Put chunks of work in a commit (the Husky program will run some checks)
+1. Run `yarn smoketest` to confirm you're not breaking anything critical
+1. Create pull request
+
+## Query string to check comparator output
+
+Use the following query string to check the comparator output: INSERT PROPER LINK HERE
```
HonkingGoose, line 73: @Belco90 Can you suggest a realistic query string we should use here?

Belco90: I tend to use this comparison for checking everything is working as expected: `?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0`

It includes major, minor, and patch semver, code of blocks, diffs, commits and PRs references, different text emphasis... I think this is a good candidate for what you are looking for.

Belco90 (approved):
Wow, fantastic addition. Thank you very much for adding these guidelines!

HonkingGoose:
It needs more work, but for now this is good enough to start with. 😄

Issue #177, opened 2021-06-26
Create automated tests


HonkingGoose:
## What feature would you like to have?

We don't run any automated test on our components, or on the website as a whole.

I think we would benefit from having some basic tests (even if it's just to confirm that a certain text actually shows on a rendered page), or that our comparator actually works.

## What problem will be solved by having the feature?

Reduce manual review work, improve confidence we're not breaking things when we update dependencies, or update our code.

## Do you want to help making the feature?

I'm willing to help out, but I don't know how get to Jest or Cypress set up for use with our Next.js app.
Also I need some "templates" or basic tests that I can then try enhancing.

Belco90:
I had this in mind and have implemented tons of tests with both Jest and Cypress for Next.js projects in the past. It was just a matter of time, but definitely I still interested in implementing them.

There are tons of examples within Next.js repo, so you can find some references there about how to set it up with Jest. This is an example with Jest + TS: https://github.com/vercel/next.js/blob/canary/examples/with-typescript-eslint-jest/README.md

HonkingGoose:
Do we still want to add more tests, or are we happy with having just the Cypress tests for now? I'm not able to work on any Jest tests, as I've not really written any new Jest tests ever.

Belco90:
I think we can consider this ticket done.

There are more things related to testing we can address in other tickets tho:

- Stubbing requests in Cypress (already has a ticket)
- Unit/integration tests with Jest + Testing Library
- Visual diff testing

HonkingGoose:
I've created issues for the remaining items, see:

- #510
- #511

Closing this issue now. 😄

PR #228, opened 2021-07-08
DO NOT MERGE: Initial cypress tests + setup

HonkingGoose:
## Changes

- Setup Cypress in a basic way
- Create basic tests

## Context

Helps with #177, but does not close it.

I'm probably not setting things up 100% correctly here. As I understand it, you have used Cypress a lot, can you maybe review the setup and suggest changes to make it better? 😄 

Also the test are probably not specific enough in some places. I'm having a hard time grabbing the correct part of the HTML to act upon. But at least we're very close to testing the "happy path" of the comparator, with a proper full fat test.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ node_modules
 .env*
 *.config*
 .eslintrc.js
+cypress/
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,3 @@
+{
+  "baseUrl": "http://localhost:3000"
+}
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
--- a/cypress/integration/about.ts
+++ b/cypress/integration/about.ts
@@ -1,0 +1,57 @@
+describe('The about page', () => {
+  it('has working link to comparator page', () => {
+    cy.visit('/')
+
+    cy.contains('Comparator').click()
+
+    cy.url().should('include', '/')
+  })
+})
+
+describe('The about page', () => {
+  it('has correct page title', () => {
+    cy.visit('/about')
+
+    cy.title().should(
+      'equal',
+      'About | Octoclairvoyant: Compare GitHub changelogs across multiple releases'
+    )
+  })
+})
+
+describe('The about page', () => {
+  it('contains images which all have a alt attribute', () => {
+    cy.visit('/about')
+    cy.get('img').each(($el) => {
+      cy.wrap($el).should('have.attr', 'alt')
+    })
+  })
+})
+
+describe('The about page', () => {
+  it('contains correct hero text', () => {
+    cy.visit('/about')
+    cy.contains('h1', 'Compare GitHub changelogs across multiple releases')
+  })
+})
+
+describe('The about page', () => {
+  it('contains h2 with features text', () => {
+    cy.visit('/about')
+    cy.contains('h2', 'Features')
+  })
+})
+
+describe('The about page', () => {
+  it('contains call to action button with correct text', () => {
+    cy.visit('/about')
+    cy.contains('button', 'Try me now!')
+  })
+})
+
+describe('The about page', () => {
+  it('contains unordered list', () => {
+    cy.visit('/about')
+    cy.contains('ul', 'Search repositories and pick a version range')
+  })
+})
--- a/cypress/integration/about.ts
+++ b/cypress/integration/about.ts
@@ -1,6 +1,6 @@
 describe('The about page', () => {
   it('has working link to comparator page', () => {
-    cy.visit('/')
+    cy.visit('/about')
 
     cy.contains('Comparator').click()
 
--- a/cypress/integration/about.ts
+++ b/cypress/integration/about.ts
@@ -4,7 +4,7 @@ describe('The about page', () => {
 
     cy.contains('Comparator').click()
 
-    cy.url().should('include', '/')
+    cy.url().should('equal', 'http://localhost:3000/')
   })
 })
 
--- a/cypress/integration/smoketests.ts
+++ b/cypress/integration/smoketests.ts
@@ -1,0 +1,61 @@
+describe('The Home Page', () => {
+  it('has a working comparator modal', () => {
+    cy.visit('/')
+
+    cy.get('[id^=downshift-0-input]')
+      .click()
+      .type('testing-library/dom-testing-library')
+    cy.wait(5000)
+    cy.contains('p', 'testing-library/dom-testing-library').click()
+
+    // TODO: make selecting first version specific to the first selector box
+
+    cy.contains('select', 'Choose a release').select('v6.16.0')
+    cy.url().should(
+      'eq',
+      'http://localhost:3000/?repo=testing-library%2Fdom-testing-library&from=v6.16.0'
+    )
+
+    // TODO: Select v8.1.0 from second version dropdown box
+  })
+})
+
+describe('Comparator smoketest', () => {
+  it('with standard query string shows expected results', () => {
+    cy.visit(
+      '/?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0'
+    )
+
+    // Confirm from and to version range are displayed
+
+    cy.contains('Changes from v6.16.0 to v8.1.0')
+
+    // Confirm versions are mentioned somewhere
+
+    cy.contains('v6.16.0')
+    cy.contains('v8.1.0')
+
+    cy.contains('h2', 'breaking changes')
+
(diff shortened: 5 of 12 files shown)


.eslintignore
```diff
@@ -5,3 +5,4 @@ node_modules
 .env*
 *.config*
 .eslintrc.js
+cypress/
```
HonkingGoose, line 8: ESLint kept complaining whenever I wanted to commit my Cypress test work, so I've shut ESLint up, by putting the whole `cypress/` folder in the `.eslintignore` file. This is probably a bad/stupid thing to do???

cypress.json
```diff
@@ -0,0 +1,3 @@
+{
+  "baseUrl": "http://localhost:3000"
```
HonkingGoose, line 2: I think it's safe to assume that we'll run the server on `localhost:3000`.
This way we don't need to put `http://localhost:3000` in our `cy.visit` commands. 😄

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
```
HonkingGoose, line 4: This is just the boilerplate Cypress file, so that we have a fixtures directory ready for use later on.

I don't know anything about using fixtures, but I think we might want to "fake" some comparator responses later on?

cypress/integration/about.ts
```diff
@@ -0,0 +1,57 @@
...
+  })
+})
+
+describe('The about page', () => {
+  it('has correct page title', () => {
+    cy.visit('/about')
+
+    cy.title().should(
+      'equal',
+      'About | Octoclairvoyant: Compare GitHub changelogs across multiple releases'
```
HonkingGoose, line 17: I tried importing the `BRIEF_DESCRIPTION` constant from the `common.ts` file but couldn't get it working.
Maybe you can try with your hot-shot TypeScript and Cypress skills? 😉 

Link to file where the constant I want to import lives: https://github.com/Belco90/octoclairvoyant/blob/main/src/common.ts

cypress/integration/about.ts
```diff
@@ -0,0 +1,57 @@
...
+    cy.get('img').each(($el) => {
+      cy.wrap($el).should('have.attr', 'alt')
+    })
+  })
+})
+
+describe('The about page', () => {
+  it('contains correct hero text', () => {
+    cy.visit('/about')
+    cy.contains('h1', 'Compare GitHub changelogs across multiple releases')
```
HonkingGoose, line 34: Would be better to import the `BRIEF_DESCRIPTION`  from `common.ts` here as well... 😄

cypress/integration/smoketests.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    // TODO: make selecting first version specific to the first selector box
+
+    cy.contains('select', 'Choose a release').select('v6.16.0')
+    cy.url().should(
+      'eq',
+      'http://localhost:3000/?repo=testing-library%2Fdom-testing-library&from=v6.16.0'
+    )
+
+    // TODO: Select v8.1.0 from second version dropdown box
```
HonkingGoose, line 19: I don't know how I can grab the second selector. I could get the first version selector to work though. 🚀 
We're very close to having the full comparator happy path tested. 😉

cypress/integration/smoketests.ts
```diff
@@ -0,0 +1,61 @@
...
+  })
+})
+
+describe('Comparator smoketest', () => {
+  it('with standard query string shows expected results', () => {
```
HonkingGoose, line 24: This test is basically trying to figure out if we show all the expected items from the standard query string.

cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["es5", "dom"],
+    "types": ["cypress", "node"]
```
HonkingGoose, line 5: I needed to add `"node"` to the `"types"` array to get VSCode to stop complaining about not finding the types it needed.
The config probably could do with a review/edit. 😉

HonkingGoose (commented):
Some things I wanted to highlight from my current work.

.eslintignore
HonkingGoose: And it turns out I'm breaking the whole build/linting run. 😄

cypress/integration/about.ts
```diff
@@ -0,0 +1,57 @@
+describe('The about page', () => {
+  it('has working link to comparator page', () => {
+    cy.visit('/')
```
HonkingGoose, line 3:
```suggestion
    cy.visit('/about')
```

.eslintignore
Belco90: We should lint cypress tests files with ESLint indeed, there is even an [ESLint plugin for Cypress](https://github.com/cypress-io/eslint-plugin-cypress) 😅

cypress.json
Belco90: Yeah, good idea.

cypress/fixtures/example.json
Belco90: Yeah, definitely useful to get some fixtures here in the future.

cypress/integration/about.ts
Belco90: This seems an issue with our current TS setup, I'll try to take a look at our `tsconfig.json` during the week.

cypress/integration/about.ts
Belco90: Aah, that's why you want to import `BRIEF_DESCRIPTION`. I disagree with important that. Better to have hardcoded values within tests 😄 Otherwise we might introduce some error to `BRIEF_DESCRIPTION` and we wouldn't realize because the test would be comparing against the same const, so you can only realize about that hardcoding the expected value in tests.

HonkingGoose:
I'll stop changing things now, as I'm only causing noise this way. 😉

cypress/integration/smoketests.ts
Belco90: I highly recommend installing [Cypress Testing Library](https://testing-library.com/docs/cypress-testing-library/intro) so you can query elements in an accessible way.

It makes it easier to get the second select by querying by `role: "select"` and `name: "To release"`

cypress/tsconfig.json
Belco90: Wouldn't be better to add this to the main `tsconfig.json`? Not really sure what's the best approach, I need to double-check it.

cypress/integration/about.ts
HonkingGoose: As we want to use a hard-coded value instead of referring to the imported `const` I'll resolve this review comment.

cypress/integration/about.ts
HonkingGoose: >  Otherwise we might introduce some error to `BRIEF_DESCRIPTION` and we wouldn't realize because the test would be comparing against the same const

Let's keep using the current hard-coded value then.

cypress/integration/smoketests.ts
HonkingGoose: > It makes it easier to get the second select by querying by `role: "select"` and `name: "To release"`

Yes that would make things way easier! 😄

cypress/tsconfig.json
HonkingGoose: For your information: I based this work on the [Cypress docs on working with TypeScript, configure tsconfig.json](https://docs.cypress.io/guides/tooling/typescript-support#Configure-tsconfig-json) section.

I'll let you figure out what the best solution is. 😉

.eslintignore
HonkingGoose: I tried setting up that ESLint for Cypress plugin locally, and that seems to work OK-ish, but not sure of the correct configuration.

I think it's probably best if you install and setup the Cypress ESLint plugin and the Cypress Testing Libary correctly yourself, and then push a new commit(s) with the new packages and configuration? I think otherwise I'll keep making unnecessary noise here. 😉 

I'm sorry for being a bit of an idiot with this whole "configuring Cypress and ESLint and Cypress Testing Libary) thing. 🙈 😄

HonkingGoose:
HI @Belco90 👋

This PR is getting conflicted with `main`. Maybe it's easiest to start again from the current `main` branch, and let you do the initial configuration/setup?

Things you can maybe do/figure out:

- [ ] Install and configure Cypress so we're not breaking existing builds/linting rules.
- [ ] Install and configure [ESLint plugin for Cypress](https://github.com/cypress-io/eslint-plugin-cypress).
- [ ] Install and configure [Cypress Testing Library](https://testing-library.com/docs/cypress-testing-library/intro).
- [ ] Figure out if the config from `cypress/tsconfig.json` should be merged with the main `tsconfig.json` file.

Things I hope to do when you're done with the above:

- [ ] Refactor tests from this branch to use the new setup
- [ ] Refactor tests to use the Cypress Testing Library to finish the comparator test suite

Belco90:
Closing in favor of #370

PR #238, opened 2021-07-10
Use sticky positioning for changelog headings

### LINKED ISSUE #147: Sticky positioning for changelog headings (breaking change/feat/fix/etc.)

Make the changelog headings like breaking change/feat/fix/miscellaneous "sticky". The idea is that when you scroll down the last heading remains in view, so you always know what category of change you're looking at.

The "sticky heading" should have a bit of vertical whitespace on the top and bottom to separate the heading title from the Octoclairvoyant header and from the content in the changelog view. 
No changelog text should be scrolling "under" the sticky header, as that will make both the heading and the changelog harder to read, and it will look bad. 😄 

Mozilla MDN Web docs has a snippet that shows the basic idea (but is lacking the vertical whitespace!):

```css
position: -webkit-sticky;
position: sticky;
top: 20px;
```

https://developer.mozilla.org/en-US/docs/Web/CSS/position has a lot of good information on using "sticky positioning", so I recommend you read that in full before starting work.

I don't think Chakra UI has a position sticky prop (yet): https://chakra-ui.com/docs/features/style-props#position so I think this will need to be done with normal CSS.

Makes it easy to see what category of changelog you're reading right now.

I have done some work on gathering the requirements and relevant documentation.
I don't think I know enough to make it work in the proper intended way, so I will let somebody else make the feature.

HonkingGoose:
## Changes

- Use `position="sticky"` and `top={0}` to make the heading sticky
- Use `bg="white"` and `py={4}` so there's whitespace separating the scrolling text and the heading

## Context

Fixes #147.

The idea here is that the changelog heading for the section you're in remains visible while scrolling.
If you reach a newer heading, the old heading is "pushed out of the way".

According to https://caniuse.com/css-sticky we do not need to use `-webkit-sticky` anymore to get this to work properly on iOS Safari. But we should test this on iOS Safari 14 to make sure before we merge this to `main`. 😉

### DIFF
--- a/src/components/RepositoryReleasesChangelog.tsx
+++ b/src/components/RepositoryReleasesChangelog.tsx
@@ -49,7 +49,15 @@ const ReleaseChangelogGroup = ({
   return (
     <Box key={title}>
       {shouldShowTitle && (
-        <Heading as="h2" size="xl" mb={4} textTransform={textTransform}>
+        <Heading
+          as="h2"
+          size="xl"
+          bg="white"
+          mb={4}
+          py={4}
+          textTransform={textTransform}
+          sx={{ position: 'sticky', top: '0px' }}
+        >
           {title}
         </Heading>
       )}
--- a/src/components/RepositoryReleasesChangelog.tsx
+++ b/src/components/RepositoryReleasesChangelog.tsx
@@ -56,7 +56,8 @@ const ReleaseChangelogGroup = ({
           mb={4}
           py={4}
           textTransform={textTransform}
-          sx={{ position: 'sticky', top: '0px' }}
+          position="sticky"
+          top={0}
         >
           {title}
         </Heading>


src/components/RepositoryReleasesChangelog.tsx
```diff
@@ -49,7 +49,15 @@ const ReleaseChangelogGroup = ({
...
       {shouldShowTitle && (
-        <Heading as="h2" size="xl" mb={4} textTransform={textTransform}>
+        <Heading
+          as="h2"
+          size="xl"
+          bg="white"
+          mb={4}
+          py={4}
+          textTransform={textTransform}
+          sx={{ position: 'sticky', top: '0px' }}
```
Belco90, line 59: [The `sx` property is considered an escape hatch](https://chakra-ui.com/docs/features/the-sx-prop#use-cases) since all basic CSS properties are available as component properties 😅. This comes from Styled System, you can read more about it [here](https://styled-system.com/).

You must set these styles through [Position props](https://chakra-ui.com/docs/features/style-props#position):

```suggestion
          position="sticky" // `pos` is also valid
          top={0}
```

Belco90 (changes requested):

src/components/RepositoryReleasesChangelog.tsx
HonkingGoose: Ah okay. I just searched through the Chakra UI repository to see if anybody had any tips for sticky positioning, and somebody showed that `sx` thing. 😄 

I agree, we should stay within the Chakra UI system as much as possible. 😄

Belco90:
I can confirm this is working fine on iOS Safari 14: https://cln.sh/IIQMBm

Belco90 (approved):

PR #245, opened 2021-07-11
Improve header component

HonkingGoose:
## Changes

- Use new colors
- Make header look more like Figma design
- Make links in header look better and have clearer hover states
- Make drawer look a bit better
- Make drawer open and close icons larger

## Context

This makes the header look more like the Figma design, though it needs a lot more work before it's 100% equal to the Figma design. So consider this a "first try" at aligning the header to the Figma design. The idea here is to get something better than the current header, that gives a "preview" of the final design.

**Known bad stuff that needs fixing**

- [x] There's some weird issue with the padding on the header box, seems the padding/size of the about/comparator links is messing with the size of the header box. (Start with full width window, resize window with mouse)
- [x] If link is not the active element, borderColor should equal header/drawer background (decided to keep the border but de-emphasize it for not active link element)
- [x] `h1` fontSize should shrink to `16px` on iPhone
- [x] `borderTop` should be `4px` when on iPhone (this is the colored bar on top of the header)

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -113,9 +113,8 @@ const Header = (props: BoxProps) => {
       as="header"
       zIndex="banner"
       {...props}
-      // TODO: borderTop should be 4px when on iPhones
-      borderTop="8px"
-      borderColor="fuchsia.700"
+      borderTop={{ base: '4px', sm: '8px' }}
+      borderColor={{ base: 'fuchsia.700', sm: 'fuchsia.700' }}
     >
       <FluidContainer py={5}>
         <Flex justify="space-between" alignItems="center">
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -114,6 +114,9 @@ const Header = (props: BoxProps) => {
       {...props}
       borderTopWidth={{ base: '4px', sm: '8px' }}
       borderColor="fuchsia.700"
+      // TODO: Remove borderBottomWidth and borderBottomColor once we fix the issue
+      borderBottomWidth="8px"
+      borderBottomColor="sky.700"
     >
       <FluidContainer py={5}>
         <Flex justify="space-between" alignItems="center">
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -44,7 +44,7 @@ const MenuLink = ({
       color="primaryTextLightmode"
       borderBottomWidth="4px"
       // TODO: If link is not the active element, borderColor should equal header/drawer background
-      borderColor={isActive ? 'fuchsia.300' : 'white'}
+      borderColor={isActive ? 'fuchsia.300' : 'none'}
       fontWeight="black"
       fontSize={{ base: '2xl', md: 'lg' }}
       _hover={{
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -43,7 +43,7 @@ const MenuLink = ({
     <RouteLink
       href={href}
       color="primaryTextLightmode"
-      borderBottomWidth="4px"
+      borderBottomWidth={isActive ? '4px' : 'none'}
       borderColor={isActive ? 'fuchsia.400' : 'fuchsia.200'}
       fontWeight="black"
       fontSize={{ base: '2xl', md: 'lg' }}
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -113,7 +113,7 @@ const Header = (props: BoxProps) => {
       zIndex="banner"
       {...props}
       borderTop={{ base: '4px', sm: '8px' }}
-      borderColor={{ base: 'fuchsia.700', sm: 'fuchsia.700' }}
+      borderColor="fuchsia.700"
     >
       <FluidContainer py={5}>
         <Flex justify="space-between" alignItems="center">
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -112,7 +112,7 @@ const Header = (props: BoxProps) => {
       as="header"
       zIndex="banner"
       {...props}
-      borderTop={{ base: '4px', sm: '8px' }}
+      borderTopWidth={{ base: '4px', sm: '8px' }}
       borderColor="fuchsia.700"
     >
       <FluidContainer py={5}>
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -131,8 +131,7 @@ const Header = (props: BoxProps) => {
               color="primaryTextLightmode"
               letterSpacing="tight"
               fontWeight="black"
-              // TODO: fontSize should be 16 px on iPhones
-              fontSize="36px"
+              fontSize={{ base: '16px', sm: '36px' }}
             >
               Octoclairvoyant
             </Heading>
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -41,13 +41,16 @@ const MenuLink = ({
   return (
     <RouteLink
       href={href}
-      color="white"
-      borderBottomWidth={isActive ? '4px' : 'none'}
-      borderColor={isActive ? 'primary.500' : 'none'}
+      color="primaryTextLightmode"
+      borderBottomWidth="4px"
+      // TODO: If link is not the active element, borderColor should equal header/drawer background
+      borderColor={isActive ? 'fuchsia.300' : 'white'}
+      fontWeight="black"
       fontSize={{ base: '2xl', md: 'lg' }}
       _hover={{
         textDecoration: 'none',
-        color: isActive ? 'none' : 'primary.300',
+        borderBottomColor: 'fuchsia.500',
+        color: isActive ? 'none' : 'primaryTextLightmode',
       }}
       width={linkWidth}
       aria-current={isActive ? 'page' : undefined}
@@ -84,14 +87,14 @@ const HeaderLinks = () => {
         aria-label="Toggle menu"
         icon={<Icon as={FaBars} />}
         variant="unstyled"
-        colorScheme="gray.50"
-        size="sm"
+        colorScheme="coolGray.900"
+        size="lg"
         onClick={onToggle}
       />
       <Drawer placement="right" isOpen={isOpen} size="full" onClose={onClose}>
         <DrawerOverlay />
-        <DrawerContent py={8} backgroundColor="gray.700">
-          <DrawerCloseButton color="gray.50" />
+        <DrawerContent py={8} backgroundColor="coolGray.100">
+          <DrawerCloseButton color="coolGray.900" size="lg" />
           <DrawerBody>
             <Flex justify="center" direction="column" h="full">
               {linksInner}
@@ -106,7 +109,14 @@ const HeaderLinks = () => {
 const Header = (props: BoxProps) => {
   const isClientSide = useIsClientSide()
   return (
-    <Box as="header" bg="gray.700" color="white" zIndex="banner" {...props}>
+    <Box
+      as="header"
+      zIndex="banner"
+      {...props}
+      // TODO: borderTop should be 4px when on iPhones
+      borderTop="8px"
+      borderColor="fuchsia.700"
+    >
       <FluidContainer py={5}>
         <Flex justify="space-between" alignItems="center">
           <Flex alignItems="center">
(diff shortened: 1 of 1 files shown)


src/components/Header.tsx
```diff
@@ -41,13 +41,16 @@ const MenuLink = ({
   return (
     <RouteLink
       href={href}
-      color="white"
-      borderBottomWidth={isActive ? '4px' : 'none'}
-      borderColor={isActive ? 'primary.500' : 'none'}
+      color="primaryTextLightmode"
+      borderBottomWidth="4px"
+      // TODO: If link is not the active element, borderColor should equal header/drawer background
```
Belco90, line 46: Won't be that achievable setting it to none?

`boderColor={isActive ? 'fuchsia.300' : 'none'}`

src/components/Header.tsx
```diff
@@ -106,7 +109,14 @@ const HeaderLinks = () => {
 const Header = (props: BoxProps) => {
   const isClientSide = useIsClientSide()
   return (
-    <Box as="header" bg="gray.700" color="white" zIndex="banner" {...props}>
+    <Box
+      as="header"
+      zIndex="banner"
+      {...props}
+      // TODO: borderTop should be 4px when on iPhones
```
Belco90, line 116: You can achieve that with [Responsive Styles using object syntax](https://chakra-ui.com/docs/features/responsive-styles#the-object-syntax):

`borderTop={{ base: '4px', md: '8px' }}`

src/components/Header.tsx
```diff
@@ -119,9 +129,11 @@ const Header = (props: BoxProps) => {
               />
             </Box>
             <Heading
+              color="primaryTextLightmode"
               letterSpacing="tight"
               fontWeight="black"
-              fontSize={{ base: 'md', md: 'xl', lg: '4xl' }}
+              // TODO: fontSize should be 16 px on iPhones
```
Belco90, line 135: Same here, you can do that with Responsive Styles through object syntax 😄

src/components/Header.tsx
```diff
@@ -41,13 +41,16 @@ const MenuLink = ({
   return (
     <RouteLink
       href={href}
-      color="white"
-      borderBottomWidth={isActive ? '4px' : 'none'}
-      borderColor={isActive ? 'primary.500' : 'none'}
+      color="primaryTextLightmode"
+      borderBottomWidth="4px"
+      // TODO: If link is not the active element, borderColor should equal header/drawer background
+      borderColor={isActive ? 'fuchsia.300' : 'white'}
```
HonkingGoose, line 47:
```suggestion
      borderColor={isActive ? 'fuchsia.300' : 'none'}
```

src/components/Header.tsx
HonkingGoose: Is that `md` breakpoint the same as the iPhone 11 size in our design?

> We've chosen to start with a design for **iPhone 11 Pro / X** which is:
> 
>     * `375px` wide
> 
>     * `812px` high

I'm bad at converting Chakra UI em units to px...

https://chakra-ui.com/docs/theming/theme#breakpoints

src/components/Header.tsx
```diff
@@ -41,13 +41,16 @@ const MenuLink = ({
   return (
     <RouteLink
       href={href}
-      color="white"
-      borderBottomWidth={isActive ? '4px' : 'none'}
-      borderColor={isActive ? 'primary.500' : 'none'}
+      color="primaryTextLightmode"
+      borderBottomWidth="4px"
+      // TODO: If link is not the active element, borderColor should equal header/drawer background
+      borderColor={isActive ? 'fuchsia.300' : 'none'}
```
HonkingGoose, line 47: I tried your suggestion, but the non-active link elements still have a visible border line below them. Seems `none` still has some color of it own???

If the link is not active, I don't want to see any border, but it should be present (due to padding concerns).

src/components/Header.tsx
HonkingGoose: Also, while we're discussing the `borderTop` property: do we want the `borderTop` to be on the `<header>` tag or on the `<body>` tag? I'm not sure which option is best.

I couldn't find the correct point to adjust the `<body>` tag in the code, so that's why I put the colored bar on the `<header>`.... 😄

src/components/Header.tsx
Belco90: Ah ok, I see what you mean now. Maybe transparent might work, I need to double-check it.

src/components/Header.tsx
Belco90: > Is that `md` breakpoint the same as the iPhone 11 size in our design?

I don't think it's _exactly_ the iPhone 11 size, but we need to define this by mobile, tablet, regular desktop, and large desktop rather than by the exact size of one device.



> Also, while we're discussing the `borderTop` property: do we want the `borderTop` to be on the `<header>` tag or on the `<body>` tag? I'm not sure which option is best.
> 
> I couldn't find the correct point to adjust the `<body>` tag in the code, so that's why I put the colored bar on the `<header>`.... 😄

Doing this in our Header is the perfect place to do so. The `body` is outside Next.js and Chakra scope, so you need to style it through [Chakra global's style](https://github.com/Belco90/octoclairvoyant/blob/main/src/customTheme.ts#L124)

src/components/Header.tsx
```diff
@@ -106,7 +109,13 @@ const HeaderLinks = () => {
 const Header = (props: BoxProps) => {
   const isClientSide = useIsClientSide()
   return (
-    <Box as="header" bg="gray.700" color="white" zIndex="banner" {...props}>
+    <Box
+      as="header"
+      zIndex="banner"
+      {...props}
+      borderTop={{ base: '4px', sm: '8px' }}
+      borderColor={{ base: 'fuchsia.700', sm: 'fuchsia.700' }}
```
HonkingGoose, line 117: I needed to change `borderColor` as well, otherwise the `borderTop` was using a black color instead of the fuchsia shade we want...

Unsure if bug in Chakra UI, or if this is just the way it needs to be when changing a property to be responsive....

src/components/Header.tsx
HonkingGoose: I've tried `transparent` and that does work, but now you cannot see that it's a link. 🙈 

I've decided to keep the borderColor, but make it clearer which is the active one and which the de-emphasized one.

src/components/Header.tsx
HonkingGoose: > I don't think it's _exactly_ the iPhone 11 size, but we need to define this by mobile, tablet, regular desktop, and large desktop rather than by the exact size of one device.

I think the `sm` breakpoint matches closest, so that's what I've used in my new commits.

> Doing this in our Header is the perfect place to do so. The `body` is outside Next.js and Chakra scope, so you need to style it through [Chakra global's style](https://github.com/Belco90/octoclairvoyant/blob/main/src/customTheme.ts#L124)

Whew, so it's in the correct place right now! 😄

src/components/Header.tsx
Belco90: If you just set `borderColor="fuchsia.700"` then you get a black border? That's weird.

HonkingGoose: I've pushed a commit that shows the bad behavior. 😄

Belco90: I've checked the issue in the preview platform. I've seen Chakra is generating another CSS class for desktop, but I've found the issue is on our side 😅 

The `borderTop` prop is a shortcut for setting border style, color, and width at once (same as `border-top` CSS property actually). Since we are passing just the width, Chakra is filling the other values with the default border style and color.

The proper way to set this then is using `borderTopWidth` instead:

```js
      borderTopWidth={{ base: '4px', sm: '8px' }}
      borderColor="fuchsia.700"
```

You can find all Chakra Border props [here](https://chakra-ui.com/docs/features/style-props#borders).

HonkingGoose: Ah we were basically taking a "shortcut" instead of using the proper separate variables. 😄

src/components/Header.tsx
```diff
@@ -106,7 +108,13 @@ const HeaderLinks = () => {
 const Header = (props: BoxProps) => {
```
HonkingGoose, line 108: Do we want to fix the "padding" problem when resizing the viewport, or do we want to leave that for the full implementation? Or are we happy to merge this and work on the padding problem later?

I think the current site has a similar problem?

Belco90: Not really sure what is this problem about.

HonkingGoose: I've just used the browser tools to measure the header total size:

| Simulated device | view port width |  view port height | `<header>` height |
| --- | --- | --- | --- |
| iPhone | 375 | 812 | 88 |
| iPad | 768 | 1024 | 83.2 |
| Desktop | 1440 | 900 | 98 |

This is really subtle, but I do notice the items below the header changing position when you resize the view port.

Belco90: You mean the filters switching from vertical stack to horizontal stack?

Belco90:
I want to take a look at

> If link is not the active element, borderColor should equal header/drawer background (decided to keep the border but de-emphasize it for not active link element)

I have the original behavior in my website, so I'll double-check what's the problem here.

src/components/Header.tsx
```diff
@@ -41,13 +41,15 @@ const MenuLink = ({
   return (
     <RouteLink
       href={href}
-      color="white"
-      borderBottomWidth={isActive ? '4px' : 'none'}
-      borderColor={isActive ? 'primary.500' : 'none'}
+      color="primaryTextLightmode"
+      borderBottomWidth="4px"
```
Belco90, line 45: I'd say this is the problem: we are now always fixing the border-bottom width to 4px, when it should be none if `isActive` is false.

Belco90 (changes requested):

src/components/Header.tsx
```diff
@@ -106,7 +108,16 @@ const HeaderLinks = () => {
...
-    <Box as="header" bg="gray.700" color="white" zIndex="banner" {...props}>
+    <Box
+      as="header"
+      zIndex="banner"
+      {...props}
+      borderTopWidth={{ base: '4px', sm: '8px' }}
+      borderColor="fuchsia.700"
+      // TODO: Remove borderBottomWidth and borderBottomColor once we fix the issue
+      borderBottomWidth="8px"
+      borderBottomColor="sky.700"
```
HonkingGoose, line 119: I've added some properties to help visualize the problem.
I've added a big line under the header.

I notice that the `sky.700` line jumps around when you resize the window.

Belco90: https://user-images.githubusercontent.com/2677072/125446140-928e8185-7a43-4480-ac3e-5d7f3de1eb49.mov

I guess this is the "problem" you mentioned? If so, that's just because the Header text is changing size between mobile, desktop, and tablet. That happened already, we can live with it for now.

If you mean something else, I'm absolutely lost.

HonkingGoose: I think the Octoclairvoyant image resize is messing with this a bit.

Ideally I'd want the baseline for the text of the Octoclairvoyant and the Comparator/About links to be on the same horizontal line.

https://en.wikipedia.org/wiki/Baseline_(typography)

---

But maybe I'm worrying too much over a temporary design...
We'll update this component later, so that it conforms to the Figma design.

I'd say we remove the debug code, review one last time, and merge it, and be done with it for now.
The header and drawer still look way better than what we have now.
That way we can focus our attention on the other items on our project board. 😉

Belco90: It might be the logo being resized too, but I would ignore it for now and fix it in a later step of the revamping process.

src/components/Header.tsx
Belco90: This is still pending, right? 😅

Belco90 (changes requested):

src/components/Header.tsx
HonkingGoose: I'll go make a commit to try your proposal. 😉

src/components/Header.tsx
```diff
@@ -42,13 +42,15 @@ const MenuLink = ({
   return (
     <RouteLink
       href={href}
-      color="white"
+      color="primaryTextLightmode"
       borderBottomWidth={isActive ? '4px' : 'none'}
```
HonkingGoose, line 46: This is what we had before for the `borderBottomWidth`. I think both options are OK for now. So I'll let you pick the one you like best. 😉

Belco90 (approved):
Lgtm! Thanks for all the alternatives you brought here.

HonkingGoose:
Whew, that took a bit of time, but we got there in the end! 😄 

The auto-merge was stuck because a conversation was not marked "resolved" yet.
I resolved the conversation to get it merged. 😉

Belco90:
> The auto-merge was stuck because a conversation was not marked "resolved" yet.
> 
> I resolved the conversation to get it merged. 😉

Yeah I assumed it was that, but it can't be managed on GitHub mobile app so I enabled auto-merged to get it done after one of us could resolve the conversations.

PR #247, opened 2021-07-12
Setup sentry

Belco90:
## Changes

- Set up Sentry for Next.js
- Additionally, Sentry integration for Vercel and GitHub have been set up too

## Context

Since we are gonna start changing and refactoring stuff, it's good to catch possible errors on production/preview environments.

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/triple-slash-reference */
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
--- a/next.config.js
+++ b/next.config.js
@@ -23,7 +23,7 @@ const SentryWebpackPluginOptions = {
   //   release, url, org, project, authToken, configFile, stripPrefix,
   //   urlPrefix, include, ignore
 
-  silent: process.env.NODE_ENV !== 'production', // Suppresses all logs in envs other than production
+  silent: !process.env.VERCEL_ENV, // Suppresses all logs if not building the app at Vercel
   deploy: {
     env: process.env.VERCEL_ENV,
   },
@@ -31,4 +31,9 @@ const SentryWebpackPluginOptions = {
   // https://github.com/getsentry/sentry-webpack-plugin#options.
 }
 
-module.exports = withSentryConfig(config, SentryWebpackPluginOptions)
+if (process.env.VERCEL) {
+  // wrap the bundle with Sentry only if built/deployed in Vercel
+  module.exports = withSentryConfig(config, SentryWebpackPluginOptions)
+} else {
+  module.exports = config
+}
--- a/next.config.js
+++ b/next.config.js
@@ -2,8 +2,9 @@
 const withBundleAnalyzer = require('@next/bundle-analyzer')({
   enabled: process.env.ANALYZE === 'true',
 })
+const { withSentryConfig } = require('@sentry/nextjs')
 
-module.exports = withBundleAnalyzer({
+const config = withBundleAnalyzer({
   async redirects() {
     return [
       {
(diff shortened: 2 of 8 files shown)


HonkingGoose:
I've never seen or used Sentry. Can you give me a high-level overview of what Sentry can do for us, or will help us with?

Belco90:
Sure! I was planning to add more context but I created the PR as a draft to check Sentry was actually working 😅 

So Sentry is basically a tool to catch and report errors generated at the production platform. You can read more about it [here](https://sentry.io/welcome/).

You can see I've introduced a new button in this PR with a callback that is trying to execute an existing method. When the user clicks that button, nothing will happen in the UI but an error is thrown in JS under the hood. Sentry will catch it and report it to us! This is the error caught by Sentry for that exploding button I introduced for testing here:

CleanShot 2021-07-12 at 19 41 02@2x

This is interesting for being aware of run-time errors and detecting when they were introduced. I don't expect this to report too many errors not to be honest, but still is useful, especially if more users start landing to the app from Renovate for example.

next-env.d.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/triple-slash-reference */
```
Belco90, line 1: This is not necessary anymore since `xo` ESLint config was removed.

next.config.js
```diff
@@ -14,3 +15,20 @@ module.exports = withBundleAnalyzer({
     ]
   },
 })
+
+const SentryWebpackPluginOptions = {
```
Belco90, line 19: The whole config has been automatically generated from `@sentry/wizard` for a Next.js project.

next.config.js
```diff
@@ -14,3 +15,20 @@ module.exports = withBundleAnalyzer({
...
+  // Additional config options for the Sentry Webpack plugin. Keep in mind that
+  // the following options are set automatically, and overriding them is not
+  // recommended:
+  //   release, url, org, project, authToken, configFile, stripPrefix,
+  //   urlPrefix, include, ignore
+
+  silent: process.env.NODE_ENV !== 'production', // Suppresses all logs in envs other than production
+  deploy: {
+    env: process.env.VERCEL_ENV,
```
Belco90, line 28: This is the only thing I added myself so we can classify errors per environment. [`VERCEL_ENV` is exposed by Vercel](https://vercel.com/docs/environment-variables#system-environment-variables) so this will be either `production`, `preview`, or `development`.

sentry.client.config.js
```diff
@@ -0,0 +1,17 @@
+// This file configures the initialization of Sentry on the browser.
+// The config you add here will be used whenever a page is visited.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs'
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN
```
Belco90, line 7: I added `NEXT_PUBLIC_SENTRY_DSN` to Vercel env vars.

next-env.d.ts
HonkingGoose: Cool! That line kept getting removed when I ran Prettier, and I wasn't sure if we needed that line or not. Saves me a bit of hassle each commit. 😄

next.config.js
HonkingGoose: We can assume that works then. 😄

next.config.js
HonkingGoose: That seems handy, you can see which environment is broken. 😄

HonkingGoose (changes requested):
The build is not passing right now, so maybe you need to fix something?

Belco90:
> The build is not passing right now, so maybe you need to fix something?

Yes, I'm on it. Basically, the command to push some files to Sentry is being executed every time `yarn build` is ran. I'm checking how to opt out of this behavior to have control over it.

next.config.js
Belco90: We are sure it's working since I forced an issue with that strange button I added right after creating the PR 😅

next.config.js
```diff
@@ -2,8 +2,9 @@
 const withBundleAnalyzer = require('@next/bundle-analyzer')({
   enabled: process.env.ANALYZE === 'true',
 })
+const { withSentryConfig } = require('@sentry/nextjs')
```
HonkingGoose, line 5:
```suggestion

const { withSentryConfig } = require('@sentry/nextjs')
```

I'd like a newline here for readability. 😉

HonkingGoose (changes requested):
Just one thing, and I think we're good to merge.

next.config.js
Belco90: I'm sorry but the spacing and sorting for the imports are set up with `eslint-plugin-import`. With our current config, it will group together (i.e. without newlines) all imports belonging to the same category. These two are external imports so they are group together:

CleanShot 2021-07-12 at 20 30 35@2x

HonkingGoose: Oh well. 😄 

Let's keep things as they are then. 😄

HonkingGoose (approved):

PR #248, opened 2021-07-12
Enable color mode

### LINKED ISSUE #240: Prepare backend so it supports light/dark colors

I have designed a light and dark variant of the Octoclairvoyant site.
I don't know if we can start using light and dark colors right now, or if we need to do more work on the backend first.

I found this documentation from Chakra UI:

https://chakra-ui.com/docs/features/color-mode

Enable us to use dark/light colors on our site.

I have the designs for light and dark mode ready.
I also designed a toggle for light/dark for desktop and mobile.

Belco90:
## Changes

- Enables color mode to switch between light and dark modes
- Hides the color mode under a feature flag named `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE` to not releasing this to production yet

If you want to enable the color mode locally, you need to:

1. create a file named `.env.local` in the root of the project (this is ignored in the repo so won't be pushed since the target of this file is to set env vars locally)
2. put `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE=true` inside this file
3. restart your local server (i.e. stop `yarn start` and rerun it again)

When we want to release this officially to the production platform it would be as simple as setting the new env var in Vercel for the production environment, and then remove the conditions checking `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE` value at some point.

## Context

Closes #240

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,5 +85,5 @@ Do the following to get a ugly button to toggle between light/dark mode:
 
 1. Create a file named `.env.local` in the root of the project (this file is on the `.gitignore` list so it won't be committed accidentally)
 1. Put `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE=true` inside the `.env.local` file.
-1. Stop the development server, and run `yarn start` to restar the server
+1. Stop the development server, and run `yarn start` to restart the server
 1. You can now use the button in the header to switch between light/dark mode
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,3 +87,36 @@ Do the following to get a ugly button to toggle between light/dark mode:
 1. Put `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE=true` inside the `.env.local` file.
 1. Stop the development server, and run `yarn start` to restart the server
 1. You can now use the button in the header to switch between light/dark mode
+
+### Setting colors for light/dark mode
+
+You only need to pass 2 values to `useColorModeValue`:
+
+- the first one is the variant for light mode
+- the second one is the variant for dark mode
+
+Save the values in a `const` that will _automagically_ get the corresponding value based on the current color mode.
+You can use this `const` anywhere, but most of the time it will be passed to a Chakra component prop.
+
+You can use `useColorModeValue` as many times you need/want inside a component to generate several variables based on color mode.
+
+### Example in code
+
+Here's an example of how to colorize a component.
+
+1. Create/edit a `const` that contains the colors.
+1. Ue the `const` within a Chakra property.
+
+```typescript
+const Footer = () => {
+  const boxBgColor = useColorModeValue('gray.50', 'gray.900')
+
+  return (
+    <Box as="footer" bg={boxBgColor}>
+      <FluidContainer></FluidContainer>
+    </Box>
+  )
+}
+
+export default Footer
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,3 +75,15 @@ Use the following query string to check the comparator output:
 ```
 ?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0
 ```
+
+## Working on dark mode locally
+
+We've hidden the color mode behind a feature flag called `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE`.
+This way we can work on the dark mode without publishing the feature.
+
+Do the following to get a ugly button to toggle between light/dark mode:
+
+1. Create a file named `.env.local` in the root of the project (this file is on the `.gitignore` list so it won't be committed accidentally)
+1. Put `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE=true` inside the `.env.local` file.
+1. Stop the development server, and run `yarn start` to restar the server
+1. You can now use the button in the header to switch between light/dark mode
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,4 +1,4 @@
-import { Box, Center, HStack, Stack } from '@chakra-ui/react'
+import { Box, Center, HStack, Stack, useColorModeValue } from '@chakra-ui/react'
 import Image from 'next/image'
 
 import poweredByVercelLogo from '@app-public/powered-by-vercel.svg'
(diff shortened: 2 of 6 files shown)


src/components/Header.tsx
```diff
@@ -66,6 +67,9 @@ const LinksStack = () => (
   >
     <MenuLink href="/">Comparator</MenuLink>
     <MenuLink href="/about">About</MenuLink>
+    {!!process.env.NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE && (
```
Belco90, line 70: This env var is already set up in Vercel for development and preview platforms.

src/components/ToggleColorModeButton.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const { colorMode, toggleColorMode } = useColorMode()
+  const isLightMode = colorMode === 'light'
+
+  return (
+    <Button
+      onClick={toggleColorMode}
+      colorScheme={isLightMode ? 'primary' : 'secondary'}
+    >
+      Toggle {isLightMode ? 'Dark' : 'Light'}
+    </Button>
```
Belco90, line 13: This button is ugly as fuck and doesn't correspond to the designs, but it's just a mechanism to allow you to force light or dark mode. It will be styled properly when releasing it officially to production.

src/customTheme.ts
```diff
@@ -99,6 +99,11 @@ const tertiaryTextDarkmode = coolGray['400']
 
 // ***** Collect constants, put them in a customTheme *****
 
+const themeConfig: ThemeConfig = {
+  initialColorMode: 'light',
+  useSystemColorMode: !!process.env.NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE,
```
Belco90, line 104: If color mode is enabled with the feature flag, then the webapp will automatically be initialized with the color mode of your OS.

Belco90:
If you want me to add an example of `useColorModeValue` hook usage to decide between different values depending on color mode enabled, let me know and I can add this to the footer.

src/components/ToggleColorModeButton.tsx
HonkingGoose: I got it working on my end with your detailed instructions. Hooray! 🎉 

That button is _nowhere near_ the final design, I'm glad we hide it from public view... 🤪 😉 

I really like the idea to put this feature behind a flag, so that we can work on it "behind the scenes", while still pushing new code to `main` instead of to a `dark_mode_branch`! 😄

CONTRIBUTING.md
```diff
@@ -75,3 +75,15 @@ Use the following query string to check the comparator output:
...
+
+We've hidden the color mode behind a feature flag called `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE`.
+This way we can work on the dark mode without publishing the feature.
+
+Do the following to get a ugly button to toggle between light/dark mode:
+
+1. Create a file named `.env.local` in the root of the project (this file is on the `.gitignore` list so it won't be committed accidentally)
+1. Put `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE=true` inside the `.env.local` file.
+1. Stop the development server, and run `yarn start` to restar the server
+1. You can now use the button in the header to switch between light/dark mode
```
HonkingGoose, line 89:
```suggestion
## Working on dark mode locally

We've hidden the color mode behind a feature flag called `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE`.
This way we can work on the dark mode without publishing the feature.

Do the following to get a ugly button to toggle between light/dark mode:

1. Create a file named `.env.local` in the root of the project (this file is on the `.gitignore` list so it won't be committed accidentally)
1. Put `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE=true` inside the `.env.local` file.
1. Stop the development server, and run `yarn start` to restart the server
1. You can now use the button in the header to switch between light/dark mode
```

HonkingGoose:
> If you want me to add an example of `useColorModeValue` hook usage to decide between different values depending on color mode enabled, let me know and I can add this to the footer.

@Belco90 I think that would be nice, as I don't understand how I'm supposed to add the colors to the components right now. 😕 We don't want to end up with a mess while I'm working on things.... 😄

src/customTheme.ts
HonkingGoose: I did see that the button changes the colors, but I guessed that was coming from the default chakra components, which supposedly all have their own darkmode styling.

Does this really matter? I intend to override all colors anyway, to make sure they're 100% according to the design... 😄

CONTRIBUTING.md
```diff
@@ -75,3 +75,15 @@ Use the following query string to check the comparator output:
 ```
 ?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0
 ```
+
+## Working on dark mode locally
```
HonkingGoose, line 79: I added some docs on how to work with the dark mode locally. 😄 

If you have anything important you want me to know, please put it in here, so I can refer to this document while I'm colorzing things... 😄

src/components/ToggleColorModeButton.tsx
Belco90: Always better to push everything to `main` and simplify the workflow. It's simple to hide things like this behind a flag.

src/customTheme.ts
Belco90: That's Chakra's default for dark mode indeed. As we discussed before, better to override them manually first, and then we swipe Chakra's default tokens for dark mode by ours.

CONTRIBUTING.md
Belco90: Nice addition! It won't be necessary eventually when we release the dark mode officially, but at that time we can convert this section into "How to enable env vars locally" I guess 😅

src/components/Footer.tsx
```diff
@@ -1,54 +1,62 @@
-import { Box, Center, HStack, Stack } from '@chakra-ui/react'
+import { Box, Center, HStack, Stack, useColorModeValue } from '@chakra-ui/react'
 import Image from 'next/image'
 
 import poweredByVercelLogo from '@app-public/powered-by-vercel.svg'
 import { REPO_URL } from '~/common'
 import FluidContainer from '~/components/FluidContainer'
 import Link from '~/components/Link'
 
-const Footer = () => (
```
Belco90, line 9: Although it seems there are tons of changes on this component, it's just a reformating after switching from arrow function with implicit return (i.e. `() => <Foo />`) to arrow function with explicit return (i.e. `() => { return <Foo /> }`)

src/components/Footer.tsx
```diff
@@ -1,54 +1,62 @@
...
-  <Box as="footer" bg="gray.50" flexShrink={0}>
-    <FluidContainer py={5}>
-      <Stack justify="space-between" direction={['column', 'row']} spacing={4}>
-        <HStack
-          shouldWrapChildren
+const Footer = () => {
+  const boxBgColor = useColorModeValue('gray.50', 'gray.900')
```
Belco90, line 10: So this is the magic line for switching between 2 values depending on color mode.

You only need to pass 2 values to `useColorModeValue`: the first one is the variant for light mode, and the second one is the variant for dark mode.

Then you save it in a const that will _automagically_ get the corresponding value based on the current color mode.

You can use that variable anywhere, but most of the time will be passed to a Chakra component prop as I did in the `Box` below.

You can use `useColorModeValue` as many times you need/want inside a component to generate several variables based on color mode.

HonkingGoose: Thank you for the example! ❤️ 

I've pushed a commit that adds your advice to the `contributing.md` file.

CONTRIBUTING.md
```diff
@@ -75,3 +75,48 @@ Use the following query string to check the comparator output:
...
+This way we can work on the dark mode without publishing the feature.
+
+Do the following to get a ugly button to toggle between light/dark mode:
+
+1. Create a file named `.env.local` in the root of the project (this file is on the `.gitignore` list so it won't be committed accidentally)
+1. Put `NEXT_PUBLIC_FEATURE_FLAG_COLOR_MODE=true` inside the `.env.local` file.
+1. Stop the development server, and run `yarn start` to restart the server
+1. You can now use the button in the header to switch between light/dark mode
+
+### Setting colors for light/dark mode
```
HonkingGoose, line 91: Can you check if the new section in the `CONTRIBUTING.md` file is correct?
For me it's easiest to have this in a file, so that I can just follow the instructions... 😄

Belco90: lgtm!

HonkingGoose (approved):

PR #260, opened 2021-07-17
Make Octoclairvoyant image link to homepage

### LINKED ISSUE #152: Clicking on Octoclairvoyant logo + text area takes you to the homepage

Clicking on the Octoclairvoyant logo and/or the Octoclairvoyant text area takes you to the homepage.

On nearly all websites clicking on their logo/brandname in the top header takes you back to their homepage, but we don't follow this pattern.

By following the convention, we conform to the end-users expectation, and make the site easier to navigate.

I don't know what's the proper way to make a link out of the logo and text area, so I'll let you do that.

If you add some basic hover, focus and active states, I can then adjust those to look nice. 😉

HonkingGoose:
## Changes

- Make Octoclairvoyant image link to homepage

## Context

Fixes #152.

This works, but I'm not sure if this is the best way of doing it. 😄 

The link is actually on the `<Box>` element that has the image in it.

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -21,16 +21,16 @@ const Header = (props: BoxProps) => {
       <FluidContainer py={5}>
         <Flex justify="space-between" alignItems="center">
           <Flex alignItems="center">
-            <RouteLink href="/">
-              <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
+            <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
+              <RouteLink href="/">
                 <Image
                   priority
                   src={mascotIcon}
-                  alt="Octoclairvoyant reading a crystal ball"
+                  alt="Go to home page"
                   placeholder="blur"
                 />
-              </Box>
-            </RouteLink>
+              </RouteLink>
+            </Box>
             <Heading
               as="h1"
               color="primaryTextLightmode"
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -21,7 +21,7 @@ const Header = (props: BoxProps) => {
       <FluidContainer py={5}>
         <Flex justify="space-between" alignItems="center">
           <Flex alignItems="center">
-            <RouteLink href="">
+            <RouteLink href="/">
               <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
                 <Image
                   priority
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -21,14 +21,16 @@ const Header = (props: BoxProps) => {
       <FluidContainer py={5}>
         <Flex justify="space-between" alignItems="center">
           <Flex alignItems="center">
-            <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
-              <Image
-                priority
-                src={mascotIcon}
-                alt="Octoclairvoyant reading a crystal ball"
-                placeholder="blur"
-              />
-            </Box>
+            <RouteLink href="/">
+              <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
+                <Image
+                  priority
+                  src={mascotIcon}
+                  alt="Octoclairvoyant reading a crystal ball"
+                  placeholder="blur"
+                />
+              </Box>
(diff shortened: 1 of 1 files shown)


src/components/Header.tsx
```diff
@@ -21,14 +21,16 @@ const Header = (props: BoxProps) => {
...
           <Flex alignItems="center">
-            <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
-              <Image
-                priority
-                src={mascotIcon}
-                alt="Octoclairvoyant reading a crystal ball"
-                placeholder="blur"
-              />
-            </Box>
+            <RouteLink href="/">
```
Belco90, line 24: The route link should be placed between the box and the image:

```js
<Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
  <RouteLink href="/">
    <Image
    // ...
```

src/components/Header.tsx
```diff
@@ -21,14 +21,16 @@ const Header = (props: BoxProps) => {
...
+              <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
+                <Image
+                  priority
+                  src={mascotIcon}
+                  alt="Octoclairvoyant reading a crystal ball"
```
Belco90, line 29: The alt of the image should reflect now this is a link to the home page:

`alt="Go to home page"`

Extracted from [Creating valid and accessible links](https://www.a11yproject.com/posts/2019-02-15-creating-valid-and-accessible-links/):
> To be a valid link it must have:
> ...
> Link content text describing where the link is going, this could be plain text **or the `alt` attribute of an image**.

Belco90 (changes requested):

src/components/Header.tsx
HonkingGoose: Good catch! I didn't know that you should use the `alt` text to describe where the link goes, when you use an image as link. 😄 Learned something new today. 😄

src/components/Header.tsx
```diff
@@ -22,12 +22,14 @@ const Header = (props: BoxProps) => {
         <Flex justify="space-between" alignItems="center">
           <Flex alignItems="center">
             <Box h={LOGO_SIZES} w={LOGO_SIZES} mr={2}>
-              <Image
-                priority
-                src={mascotIcon}
-                alt="Octoclairvoyant reading a crystal ball"
-                placeholder="blur"
-              />
+              <RouteLink href="/">
```
HonkingGoose, line 25: The `RouteLink` component has a really weird outline... The outline does not cover the full box of the image now. You can see this by refreshing the page and using <kbd>Tab</kbd> to tab into the image.

Belco90 (dismissed):

src/components/Header.tsx
Belco90: Let me check this.

Belco90: I don't see any outline in this component, it's not focused on tab pressed 😕

HonkingGoose: bad_focus_outline

This picture shows the bad focus outline on the h1 image in Firefox 90.

Also, we have a problem with the background, see the white bar when the page is on fullscreen mode. I would have expected the light gray color to extend all the way to the footer.

Belco90: If you put the Box outside the RouterLink it gets fixed? I've seen it's semantically ok to have it like that, so if it works that way let's revert it.

About the background color, I'd say that's out of the scope of this PR, but definitely has to be fixed.

HonkingGoose: > background color, I'd say that's out of the scope of this PR, but definitely has to be fixed.

It's out of scope, but wanted to mention it anyways. 😄 

I'll go make an issue to track this bug. 😉

HonkingGoose: > If you put the Box outside the RouterLink it gets fixed? I've seen it's semantically ok to have it like that, so if it works that way let's revert it.

I'll go try, see if it fixes the bad focus outline. 😄

HonkingGoose: fix-bad-outline-commit-6dcb39a

This is what we have now, with the new commit. The focus outline behaves a lot better now. 😄

Belco90 (approved):

PR #261, opened 2021-07-17
Check forbidden branches in pre-commit hook

Belco90:
## Changes

Since I have an admin role, I can push to `main` without a problem. This is leading me to push without realizing I'm working in the default branch.

I'm including this small pre-commit hook to prevent committing to `main` branch so I can move changes to the corresponding branch to follow the PR workflow. It can be bypassed with `-n` option if I need to push directly to `main` for whatever reason.

## Context

 There wasn't specific ticket for this.


Belco90:
The proof this works:
  
CleanShot 2021-07-17 at 11 52 00@2x

.husky/forbidden-branches.sh
```diff
@@ -0,0 +1,10 @@
+# shellcheck disable=SC2006
+BRANCH=`git rev-parse --abbrev-ref HEAD`
```
HonkingGoose, line 2:
```suggestion
BRANCH=$(git rev-parse --abbrev-ref HEAD)
```

I ran the shell script through the shellcheck program, it suggests we use this syntax. I guess there's a reason you use the backtick syntax?

HonkingGoose (commented):
I like this idea a lot! 😄

One question though:

.husky/forbidden-branches.sh
Belco90: You are right! My IDE was warning me about it, but when I told it to fix it, it wasn't changed 🤔 

Let me fix this.

Belco90: Done! I've moved the script outside husky path since files there are supposed to match a git hook.

HonkingGoose:
I checked out this PR branch locally, and I tried to commit on top of this branch, and Husky complains:

```bash
.husky/pre-commit: line 8: .: forbidden-branches.sh: file not found
husky - pre-commit hook exited with code 1 (error)
```

I'm using Git for Windows, you use ZSH, so maybe there's an incompatibility here?

Belco90:
> I checked out this PR branch locally, and I tried to commit on top of this branch, and Husky complains:
> 
> ```shell
> .husky/pre-commit: line 8: .: forbidden-branches.sh: file not found
> husky - pre-commit hook exited with code 1 (error)
> ```
> 
> I'm using Git for Windows, you use ZSH, so maybe there's an incompatibility here?

I've moved the script within pre-commit file, can you give it a try again? Thanks for double-checking!

HonkingGoose:
> I've moved the script within pre-commit file, can you give it a try again? Thanks for double-checking!

That seems to have done the trick, I can commit on this PR branch now.
Not sure if the script will actually work to prevent me committing to `main`, I think the only way to find out is to merge this PR... 😄

Belco90:
> Not sure if the script will actually work to prevent me committing to `main`, I think the only way to find out is to merge this PR... 😄

I checked it before pushing, it's still working!

.husky/pre-commit
```diff
@@ -5,3 +5,14 @@
 [ -n "$CI" ] && exit 0
 
 yarn lint-staged
+
+# Prevent commit if default branch
```
HonkingGoose, line 9:
```suggestion
# Prevent commit if on main branch
```

I think this comment is more accurate?

HonkingGoose (changes requested):

HonkingGoose (approved):

HonkingGoose:
I confirm this works properly for me now as well. I cannot commit to `main` without Husky warning me about it.

PR #272, opened 2021-07-20
Pin GitHub Action digests to specific commit hash

HonkingGoose:
## Changes

- Pin GitHub Action digests to specific commit hash

## Context

Tag based versioning has a weakness: maintainers can change the commit to which the tag points.
By pinning to the commit hash, we have full control over which version of the action we run.

Read the [GitHub docs, security hardening for github actions, using third party actions](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions), section on "Pin actions to a full length commit SHA".

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,5 @@
 {
-  "extends": ["config:base"],
+  "extends": ["config:base", "helpers:pinGitHubActionDigests"],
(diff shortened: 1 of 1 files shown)


Belco90 (approved):
Another TIL for me!

PR #278, opened 2021-07-24
Improve label text for comparator

HonkingGoose:
## Changes

- Change label text for input fields so that it matches the Figma wording

## Context

The idea here is to get some small improvements from the Figma design into our app.
Both components still need a big overhaul, so all of this is just temporary until we can implement the full design. 😉


src/components/ReleaseVersionsRangeFormControl.tsx
```diff
@@ -45,14 +45,12 @@ const ReleaseVersionsRangeFormControl = (props: StackProps) => {
   const [fromReleases, toReleases] = mapReleasesRange(releases)
 
   const selectPlaceholder =
-    Array.isArray(releases) && releases.length === 0
-      ? 'Releases not found'
-      : 'Choose a release'
+    Array.isArray(releases) && releases.length === 0 ? 'No releases found' : ''
```
HonkingGoose, line 48: I noticed when there's no releases found, we show a error in the version select dropdown.

The dropdown box does not have enough space to explain why there could be no version, or what the user can do to troubleshoot:

- Check on GitHub if the repository really has no releases -> no problem.
- Check on GitHub and repository does have a release(s) -> report bug to Octoclairvoyant.

I think it would be nice to show a bigger popup box/modal, that shows a proper error message, and has helpful text on how to debug/report problems, etc.

Do you want an issue to discuss this idea in more detail?

src/components/RepositorySearchCombobox.tsx
```diff
@@ -17,7 +17,7 @@ import { useCombobox } from 'downshift'
 import debounce from 'lodash/debounce'
 import * as React from 'react'
 import { useEffect, useMemo, useState } from 'react'
-import { FaArrowDown, FaArrowUp } from 'react-icons/fa'
+import { FaSearch, FaArrowUp } from 'react-icons/fa'
```
HonkingGoose, line 20: This is the closest match to the Figma design that I could find in the current icon set. These icons should be replaced with the Heroicon icons sometime later.

At least now we use a logical icon for the default state. 😉 

I've kept the up arrow when you slide the drawer back to the top.

Belco90: Sorry but I have to disagree finally with this icon change.

This button won't search anything when pressed, but just open the select dropdown. This is even reflected in the component accessibility:

element attribute showing it's just a toggling button for the select

So the search icon doesn't represent its behavior properly. We need to discard this change, but we can probably include the button as part of the input itself. I didn't do this originally to make clear to the user they can type within the input, but that field is just a combo box so we can put the toggle button within it on the right side.

What do you think?

src/components/ReleaseVersionsRangeFormControl.tsx
Belco90: What about displaying the reason why the comparison is empty where we display the results?

Belco90: Elaborating a bit more on this: I prefer to avoid showing modals if the user workflow is not blocked anyhow.

For example, it makes sense to show a modal blocking the page when the GitHub requests limit is exceeded because the user can't use the website without authorization.

But in this case, it's just an explanation about why there are no results, so the user workflow is not blocked by anything, hence the preference on avoid a modal.

HonkingGoose: > What about displaying the reason why the comparison is empty where we display the results?

I don't know if we have room for a big error message in that dropdown box? It just about manages to fit the current text in there... 😉 

> But in this case, it's just an explanation about why there are no results, so the user workflow is not blocked by anything, hence the preference on avoid a modal.

Ah okay, I had not considered that this is a non-blocking error, and so should not display a big fat modal... 😉

Belco90: > I don't know if we have room for a big error message in that dropdown box? It just about manages to fit the current text in there... 😉

I mean leaving the dropdown as it is and displaying the explanation you mentioned in the results area:
imagen

src/components/RepositorySearchCombobox.tsx
HonkingGoose: > So the search icon doesn't represent its behavior properly. We need to discard this change,

I've discarded this change with a `git revert`. 😉 

>  but we can probably include the button as part of the input itself. I didn't do this originally to make clear to the user they can type within the input, but that field is just a combo box so we can put the toggle button within it on the right side.
> 
> What do you think?

The Figma design has a button with a search icon, that's next to the text input field.
I don't know if the Figma design is what you actually want to build?

Belco90: > The Figma design has a button with a search icon, that's next to the text input field.
I don't know if the Figma design is what you actually want to build?

The other way around: update the design to fit this behavior. I didn't think about this was a combo box with autocomplete, not a pure search. My bad 😓

src/components/ReleaseVersionsRangeFormControl.tsx
HonkingGoose: > > I don't know if we have room for a big error message in that dropdown box? It just about manages to fit the current text in there... 😉
> 
> I mean leaving the dropdown as it is and displaying the explanation you mentioned in the results area:
> imagen

I like your idea! 😄 It gives us the room to explain what went wrong, without blocking the user from re-trying with a different repo name.

HonkingGoose: I've opened issue #282 to discuss this more if needed. 😉

src/components/RepositorySearchCombobox.tsx
HonkingGoose: > The other way around: update the design to fit this behavior. I didn't think about this was a combo box with autocomplete, not a pure search. My bad 😓

I was wrong as well, so don't feel too bad... 😉 

We'll discuss the proper design on Figma together, as that is more visual than GitHub. 😄

src/components/RepositorySearchCombobox.tsx
```diff
@@ -97,7 +97,7 @@ const RepositorySearchCombobox = ({
       position="relative"
       {...getComboboxProps()}
     >
-      <FormLabel {...getLabelProps()}>Repository</FormLabel>
+      <FormLabel {...getLabelProps()}>Type repository name</FormLabel>
```
Belco90, line 100: I'd suggest saying "enter" since you can paste the repo name for example.

```suggestion
      <FormLabel {...getLabelProps()}>Enter repository name</FormLabel>
```

Belco90 (changes requested):

src/components/RepositorySearchCombobox.tsx
HonkingGoose: Let's go with `enter` then. 😉

Belco90 (approved):

HonkingGoose:
@Belco90 There are still 2 conversations you need to resolve. 😉

PR #315, opened 2021-08-08
Improve blockquote colors

HonkingGoose:
## Changes

- Use new colors for the `BlockQuote.tsx` component

## Context

Steps to test/review:

1. Use our standard comparison string: `?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0`.
1. Search for string `This should get you going on the upgrade`, and scroll to the blockquote just below this string.
1. Edit the HTML of the blockquote with the browser developer tools, so that you remove the `a` tags, and put in some placeholder text that's large enough to result in multiple lines.
1. Now you can review the styling.

### DIFF
--- a/src/components/BlockQuote.tsx
+++ b/src/components/BlockQuote.tsx
@@ -5,8 +5,8 @@ const BlockQuote = (props: BoxProps) => (
     as="blockquote"
     px={4}
     borderLeftWidth={4}
-    borderLeftColor="gray.300"
-    color="gray.500"
+    borderLeftColor="coolGray.200"
+    color="tertiaryTextLightmode"
     {...props}
   />
 )


Belco90 (approved):
I followed the testing steps and lgtm!

HonkingGoose:
> I followed the testing steps and lgtm!

It's really tricky testing things that are not part of the standard query string.

I was thinking maybe we can create our own "fixed input" which has all the required fields/data and such? I don't know if we can create a file in the repository and load that in the testing environment. I think this is called a "fixture" in testing lingo... The standard query string could serve as a boilerplate, that we then adjust to our liking, so that it contains all components.

Or maybe you have an even better idea? 😄

PR #378, opened 2021-09-13
Migrate old Cypress tests to use Testing Library features

HonkingGoose:
## Changes

- Migrate old Cypress test so that I'm using Testing Library features more often

## Todo's:

- [ ] Get review by @Belco90 
- [ ] Implement review feedback/tips and tricks
- [ ] Remove comments that are no longer needed
- [ ] Final review and merge

Closes issue #373.

Marking this as a draft PR, as I need feedback and help! 😉

### DIFF
--- a/cypress/integration/comparator.spec.ts
+++ b/cypress/integration/comparator.spec.ts
@@ -1,2 +1,49 @@
 // TODO: implement e2e for comparator here
+
+// Write new test that fills out the form so that it ends up with the standard comparator query, to confirm our form is functional.
+
+// Ask Belco how to get cy.findByText(/Changes from v6.16.0 to v8.1.0/i).should('exist') to wait for the result.
+// We should mock/fixture the result somehow, as we're expecting the same results each time... and it's really easy to hit the API limits unauthenticated.
+// For now these tests do work properly, but they're not using any Testing Library queries.
+
+it('should show expected results when using standard query string', () => {
+  cy.visit(
+    '/?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0'
+  )
+
+  // Confirm from and to version range are displayed
+
+  cy.contains('Changes from v6.16.0 to v8.1.0')
+
+  // Confirm versions are mentioned somewhere
+
+  cy.contains('v6.16.0')
+  cy.contains('v8.1.0')
+
+  cy.contains('h2', 'breaking changes')
+
+  cy.contains('span', 'v7.0.0')
+
+  cy.contains('h5', 'Drop Node 8')
+  cy.contains('p', 'Node 10 or greater is required.')
+  cy.contains('a', 'out of LTS')
+  cy.contains('a', '#459')
+  cy.contains('a', 'c3ab843')
+  cy.contains('a', '#430')
+
+  cy.contains('pre', 'devDependencies')
+
+  cy.contains('h2', 'features')
+
+  cy.contains('h2', 'bug fixes')
+
+  cy.contains('h2', 'reverts')
+
+  cy.contains('h2', 'recommendations')
+
+  cy.contains('h2', 'chore')
+})
+
+// Keep the export {} braces here!
+
 export {}
--- a/cypress/integration/home.spec.ts
+++ b/cypress/integration/home.spec.ts
@@ -1,3 +1,31 @@
+it('should show correct page title', () => {
+  cy.visit('/')
+
+  cy.title().should(
+    'equal',
+    'Homepage | Octoclairvoyant: Compare GitHub changelogs across multiple releases'
+  )
+})
+
+it('should display h1 heading with correct text', () => {
+  cy.visit('/')
+
+  cy.findByRole('heading', { level: 1, name: 'Octoclairvoyant' }).should(
+    'exist'
+  )
+})
+
+it('should display h2 heading with correct text', () => {
+  cy.visit('/')
+
+  cy.findByRole('heading', {
+    level: 2,
+    name: 'Compare GitHub changelogs across multiple releases',
+  }).should('exist')
+})
+
+// TODO: Write test that checks for image being present (alt text is already checked via some other linter).
+
 it('should have a working link to comparator page', () => {
   cy.visit('/')
 
@@ -5,5 +33,16 @@ it('should have a working link to comparator page', () => {
 
   cy.url().should('equal', 'http://localhost:3000/comparator')
 })
+
+// Ask Belco how to get the footer with .findByRole, I can't get it working with footer or with contentinfo.
+
+it('should display footer text', () => {
+  cy.visit('/')
+
+  cy.contains('footer', 'Created with')
+  cy.contains('footer', 'by Mario')
+})
+
+// NOTE: Keep the export {} function below, do not remove it!
 
 export {}


cypress/integration/comparator.spec.ts
```diff
@@ -1,2 +1,49 @@
...
+// For now these tests do work properly, but they're not using any Testing Library queries.
+
+it('should show expected results when using standard query string', () => {
+  cy.visit(
+    '/?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0'
+  )
+
+  // Confirm from and to version range are displayed
+
+  cy.contains('Changes from v6.16.0 to v8.1.0')
```
HonkingGoose, line 16: I tried replacing this with:

```suggestion
  cy.findByText(/Changes from v6.16.0 to v8.1.0/i).should('exist')
```

But it doesn't work because the `cy.findByText` function does not wait until we get the response from the real API... 😄 
Not sure how I tell Cypress/Testing Library to wait until we get a response before checking any of the conditions in the test.

cypress/integration/comparator.spec.ts
```diff
@@ -1,2 +1,49 @@
 // TODO: implement e2e for comparator here
+
+// Write new test that fills out the form so that it ends up with the standard comparator query, to confirm our form is functional.
+
+// Ask Belco how to get cy.findByText(/Changes from v6.16.0 to v8.1.0/i).should('exist') to wait for the result.
+// We should mock/fixture the result somehow, as we're expecting the same results each time... and it's really easy to hit the API limits unauthenticated.
+// For now these tests do work properly, but they're not using any Testing Library queries.
+
+it('should show expected results when using standard query string', () => {
+  cy.visit(
```
HonkingGoose, line 10: I think we should have a fixture/mock of the real response, that we then use to test with.
I'm hitting the API limits within 5 minutes of starting my work on the tests, because cypress auto-runs whenever you change a test... 😄

cypress/integration/home.spec.ts
```diff
@@ -6,4 +34,15 @@ it('should have a working link to comparator page', () => {
   cy.url().should('equal', 'http://localhost:3000/comparator')
 })
 
+// Ask Belco how to get the footer with .findByRole, I can't get it working with footer or with contentinfo.
+
+it('should display footer text', () => {
+  cy.visit('/')
+
+  cy.contains('footer', 'Created with')
+  cy.contains('footer', 'by Mario')
```
HonkingGoose, line 43:
```ts
  cy.contains('footer', 'Created with')
  cy.contains('footer', 'by Mario')
```

How should I change this so that it uses `cy.findByRole`?
I tried `cy.findByRole('footer')` but that doesn't work... 😄

HonkingGoose (commented):
Some things I need help with! 😉

cypress/integration/comparator.spec.ts
Belco90: Yes, we definitely need to stub all requests to GitHub in our Cypress tests. Otherwise, we can hit the API limit in the CI and get failing tests.

This is the official way Cypress provides to do so: https://docs.cypress.io/guides/guides/network-requests#Stubbing

It's ok for now to stub those requests per test basics with hardcoded values, and then we can extract some generic fixtures.

cypress/integration/comparator.spec.ts
Belco90: When we have stubbed the requests, then we can wait for them to be completed like this: https://docs.cypress.io/guides/guides/network-requests#Waiting

cypress/integration/comparator.spec.ts
```diff
@@ -1,2 +1,49 @@
...
+
+  cy.contains('h2', 'features')
+
+  cy.contains('h2', 'bug fixes')
+
+  cy.contains('h2', 'reverts')
+
+  cy.contains('h2', 'recommendations')
+
+  cy.contains('h2', 'chore')
```
Belco90, line 44: All these should be checked with `findByRole` which is not tied to implementation details.

cypress/integration/home.spec.ts
```diff
@@ -1,3 +1,40 @@
...
+})
+
+it('should display h3 heading with correct text', () => {
+  cy.visit('/')
+
+  cy.findByRole('heading', {
+    level: 3,
+    name: 'Features',
+  }).should('exist')
+})
```
Belco90, line 34: It's better to wrap all this in a single test. The philosophy should be trying to wrap similar assertions under the same test, so these can be grouped in a `"should display corresponding information"` test where you visit the home page just once and then assert expected elements.

Creating a test for each one of those assertions will make our test execution really slow since Cypress has to tear down and set up the tests again, visiting the home page over and over.

cypress/integration/home.spec.ts
Belco90: I highly recommend using https://testing-playground.com/ to find out what queries are available to get something on an existing website.  You can install the Chrome/Firefox extension so then you can inspect https://octoclairvoyant.vercel.app/ like this:

CleanShot 2021-09-13 at 17 12 45@2x

It seems there is no accessible way to check the footer by its role, so we would have to stick to the current `cy.contains` here.

cypress/integration/home.spec.ts
```diff
@@ -1,3 +1,40 @@
...
+
+// TODO: Write test that checks for image being present (alt text is already checked via some other linter).
```
Belco90, line 36: You should use `ByAltText` for getting images: https://testing-library.com/docs/queries/byalttext

Once you have it, you can check that the `src` attribute has the desired value.

Belco90 (changes requested):

cypress/integration/home.spec.ts
HonkingGoose: That plugin is really cool! 😄

cypress/integration/home.spec.ts
HonkingGoose: Wrapped most of the tests under a single test now. Should be faster to run now. 😄

cypress/integration/comparator.spec.ts
Belco90: I can help you with this one if you don't mind me pushing code to your branch.

Just let me know which repository and versions you'd like to use for the test and I can do that in this PR.

HonkingGoose: > I can help you with this one if you don't mind me pushing code to your branch.

Yes, please help me! I've never done this before. I've read the docs, but it's still not clear how I should start.

Feel free to push code to the PR branch, just don't force-push anything. 😉 
I'm not available for the rest of the day, so feel free to work on this, you won't get any new stuff from me today. 😄 

Might be good if you could write down the steps you're doing, so that I can learn how this works, I don't know if seeing the code will be good enough for me to understand the steps. 😉

cypress/integration/comparator.spec.ts
HonkingGoose: Might as well give me an example of telling Cypress to wait as well, if you're going to help with this as well! 😄

cypress/integration/comparator.spec.ts
HonkingGoose: Okay! I'll do that after the stubbing and waiting part is done by you.

cypress/integration/home.spec.ts
```diff
@@ -1,3 +1,37 @@
...
+  )
+
+  cy.findByRole('heading', {
+    level: 2,
+    name: 'Compare GitHub changelogs across multiple releases',
+  }).should('exist')
+
+  // TODO: check 'src' property of the found img tag is correct
+  // NOTE: Testing playground suggests: cy.findByRole('img', { name: /octoclairvoyant reading a crystal ball/i })
+  cy.findByAltText(/octoclairvoyant reading a crystal ball/i).should('exist')
```
HonkingGoose, line 24: Not sure how to drill down to get the `src` property out of the found element....

cypress/integration/comparator.spec.ts
Belco90: I'll do it then. Not sure if I'll be able to do so today, but during the week for sure.

HonkingGoose: > Just let me know which repository and versions you'd like to use for the test and I can do that in this PR.

I think doing the standard query string would be good, as that has most of the styling we care about in the response. 😉 

```
?repo=testing-library%2Fdom-testing-library&from=v6.16.0&to=v8.1.0
```

Belco90: I had an idea so we have a workaround until I can set the stubs properly (I want to use msw.js and I need to check a couple of things).

I've pushed some changes to this branch so you can use a personal GitHub access token for running the tests in Cypress! I added some instructions to the `CONTRIBUTING.md` file. The CI is already set up with a personal token I generated so it's already running with it, getting an extended rate limit.

Check it out and let me know what you think!

HonkingGoose: At least now we can use a token to increase the amount of API calls. 😄

Belco90:
Hey @HonkingGoose! Will you be able to make some progress on this? I can carry on with the remaining changes if not.

HonkingGoose:
> Hey @HonkingGoose! Will you be able to make some progress on this? I can carry on with the remaining changes if not.

I really want to get this PR finished sometime soon! The problem is that I have no clue how to get started with any of these things we wanted:

- Stub request per test basics with hardcoded values
- How to tell Cypress to wait until we get a response from GitHub
- How to drill down to the `src` property of a found `img` element

Can you help me get started with this?

Belco90:
>     * Stub request per test basics with hardcoded values
>
Don't worry about this one for now. With the workaround I added for using a token when testing, it should be fine to target the real API for now.

--- 
>     * How to tell Cypress to wait until we get a response from GitHub
> 
I'm afraid until we stub requests we can't address this properly, so the only thing we can do is to wait a few ms. [`eslint-plugin-cypress` might complain about waiting for an arbitrary time](https://github.com/cypress-io/eslint-plugin-cypress/blob/master/docs/rules/no-unnecessary-waiting.md) but we have to do it for now, so just suppress the corresponding rule for the error line.

---
>     * How to drill down to the `src` property of a found `img` element
> 
IIRC this should be doable using `.should('have.attr', 'src', 'expected-src.jpg')`


Let me know if you need help with anything else!

HonkingGoose:
Ready for a review on the work in progress, this should be a lot better than what we had at first, and it properly passes the ESLint tests now. 😄

PR #464, opened 2021-10-19
Make call to action button component

### LINKED ISSUE #269: Make call to action button component

I'd like to have a component for the call to action button.

The button should match the Figma design `1:1`.

This will allow us to use the same style of button in multiple places.
The button will also have the colors for light and dark mode baked in.

Yes

I'm willing to help with this, I already made the CTA button design on Figma.

I really don't understand how Chakra UI wants me to define the component and how to import it properly in other pages. So I'll let @Belco90 do the component, and I hope that I can learn from their PR how to do this myself. 😄

Belco90:
## Changes

- Implements new `CTAButton` component
- Uses new component homepage

## Context

Closes #269 

Not sure if the hover and active status look great so lighter?

sizzy-photo-studio-2021-10-19T180302 755Z

 

https://user-images.githubusercontent.com/2677072/137966673-62b9dbf0-3400-49ed-a946-0f64365873b2.mp4


Belco90:
> Maybe we can do something with the box shadow around the button when it's clicked/active? That would retain the shape of the button, without losing text contrast.

Sure, I can do that.

> The problem with having fuchsia as a text color, is that it needs to be a really dark fuchsia on a white background, so I cannot do anything with the call to action text itself, like lighten it, because that would lead to unreadable contrast.

Do we actually need to stick to fuchsia as text color? It would be probably better to leave the background in fuchsia and the text color in white, so it improves the contrast and we don't need so lighter colors for hovering/clicking. Something closer to the Toggle Dark button, but following your design.

Issue #258, opened 2021-07-16
Custom 404 page


HonkingGoose:
### What feature would you like to have?

I just noticed our 404 page is really generic, and doesn't fit with our branding.
I'd like to at least customize the font/text style in the 404 page.

Even better would be to have use our mascot Octoclairvoyant with a suitable text under it:

"Octoclairvoyant cannot divine this page" or something like that. A bit of fun maybe. 😉 

Even better would be a nice custom pose/expression on the Octoclairvoyant macot, but that's probably out of scope.

### What problem will this feature solve?

Ensure 404 page is thematically the same as the rest of our pages.
Inject some personality into the page.

### Do you want to help make the feature?

Yes

### Is there anything else we need to know?

I don't how how hard/easy is it to customize the 404 page.
Next.js does support making a custom 404 page: https://nextjs.org/docs/advanced-features/custom-error-page#404-page

We need to discuss what kind of scope we want for the 404 page first!
I'm willing to design a new 404 page based on the scope we decide on together. 😉

Belco90:
> Even better would be to have use our mascot Octoclairvoyant with a suitable text under it:
> 
> "Octoclairvoyant cannot divine this page" or something like that. A bit of fun maybe. 😉

I love this idea!

> Even better would be a nice custom pose/expression on the Octoclairvoyant macot, but that's probably out of scope.

This seems out of the scope indeed, but I have the original Adobe Illustrator file. Could we do something with that?

> I don't how how hard/easy is it to customize the 404 page.
> Next.js does support making a custom 404 page: https://nextjs.org/docs/advanced-features/custom-error-page#404-page
> 
> We need to discuss what kind of scope we want for the 404 page first!
> I'm willing to design a new 404 page based on the scope we decide on together. 😉

404 pages are really easy to set up in Next.js, it's just a matter to include a `404.ts` within `pages` as you saw in the documentation.

Regarding the scope for 404, I'd say that's pretty clear right? So we render it when accessing pages other than `/` or `/comparator`.

HonkingGoose:
> 404 pages are really easy to set up in Next.js, it's just a matter to include a `404.ts` within `pages` as you saw in the documentation.

I bashed something together myself on a throwaway branch. Making the 404 page itself was quite easy, I just copy pasted most of the stuff we already use for our homepage right now. That got me most of the way there with regards to text/layout.

[...]

PR #512, opened 2021-11-01
Add cancel workflow action before each step

Belco90:
## Changes

- run `cancel-workflow-action` at the beginning of each step

## Context

There wasn't an issue for this. I just realized we should cancel each workflow action in case there are several pipelines running for the same PR. Since only the last one is the one we are interested in, we will cancel the stale ones to save GitHub Actions minutes and resources.


HonkingGoose:
Maybe we can use the GitHub-native `concurrency` property to achieve the same goal without using another action which needs `write` rights?

https://docs.github.com/en/actions/deployment/about-deployments/deploying-with-github-actions#using-concurrency

Belco90:
> Maybe we can use the GitHub-native `concurrency` property to achieve the same goal without using another action which needs `write` rights?
> 
> https://docs.github.com/en/actions/deployment/about-deployments/deploying-with-github-actions#using-concurrency

I don't think we can use the `concurrency` property for this. The idea of that property is to limit workflows to a single instance at the same time. This makes sense for a deployment workflow, where you only want one single workflow to be running at a time.

Let me explain this with an example. Assuming we have PR-1 and PR-2 open:

- we commit A and B to PR-1 sequentially in time
- we commit C and D to PR-2 sequentially in time

This would trigger 4 pipelines:

1. CI workflow: PR-1_commit-A
2. CI workflow: PR-1_commit-B
3. CI workflow: PR-2_commit-C
4. CI workflow: PR-2_commit-D

These 4 workflows would be run in parallel by default at the same time: `1 | 2 | 3 | 4`. Those 4 workflows are the same type: CI workflow. What we want is to keep just 2) and 4), and cancel the others since are stale workflows.

With the `concurrency` property, all 4 workflows would be still run, but one at a time rather than in parallel: `1 --> 2 --> 3 --> 4`. This is not what we want at all. There is a `cancel-in-progress` option, but that would end up in only 4) running since all previous three would be canceled.

With the `cancel-workflow-action` action, we only run 2) and 4) in parallel since 1) and 3) are canceled: `2 | 4`

So I'm afraid that, if I understood that documentation correctly, we need to stick to `cancel-workflow-action`, at least for now, since it has a different purpose than `concurrency` native property.

Belco90:
Btw, I tried to extract the common steps with [reusable workflows](https://github.blog/changelog/2021-10-05-github-actions-dry-your-github-actions-configuration-by-reusing-workflows/) but doesn't seem to work as expected.

HonkingGoose:
> With the `concurrency` property, all 4 workflows would be still run, but one at a time rather than in parallel: `1 --> 2 --> 3 --> 4`. This is not what we want at all. There is a `cancel-in-progress` option, but that would end up in only 4) running since all previous three would be canceled.

Oh that's not what we want. Thank you for the explanation on _why_ it won't work for us right now at least! ❤️ 

I hope GitHub will build a native "stop wasting CI minutes on stale stuff" feature into their product someday.

I'm still not wild about the idea of adding a new action to our workflow files though, but I guess we're out of other options for now?

> Btw, I tried to extract the common steps with [reusable workflows](https://github.blog/changelog/2021-10-05-github-actions-dry-your-github-actions-configuration-by-reusing-workflows/) but doesn't seem to work as expected.

The whole re-usable workflow thing is quite tricky!

Belco90:
> I'm still not wild about the idea of adding a new action to our workflow files though, but I guess we're out of other options for now?

I'm afraid so. I have checked some important repos and all are using that cancel workflow action, so there is no built-in feature to do something similar for now.

> The whole re-usable workflow thing is quite tricky!

It is! I'd like to have anchors and extensible hidden jobs as GitLab does to be able to extract common parts for each job. I've subscribed to [this topic](https://github.community/t/support-for-yaml-anchors/16128) to keep an eye on such features.

.github/workflows/ci.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Arrange
+      - name: Cancel Previous Runs
```
HonkingGoose, line 16: I think we need to do more config to allow the cancel action to work with forks? [^cancel-fork]

We should also restrict the permissions for the action as much as possible, so it only uses what it really needs, and no more. [^cancel-permissions]

[^cancel-fork]: https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks

[^cancel-permissions]: https://github.com/styfle/cancel-workflow-action#advanced-token-permissions

HonkingGoose (changes requested):
I think we still need some changes?

.github/workflows/ci.yml
Belco90: I don't think we need to worry about forks since those are run outside our repo/org.

I'll set the proper token permission tho.

Belco90: I've set the read-only permission as explained in their docs:

imagen

And set the `actions: write` in the workflow yml file.

HonkingGoose: Okay so the repository defaults to `read` permission, unless we elevate it manually in the workflow `.yml` files. That's a good security posture to have as a baseline.

.github/workflows/ci.yml
```diff
@@ -6,13 +6,16 @@ on:
 
 permissions:
   contents: read
+  actions: write
```
HonkingGoose, line 9: Maybe we should put the `actions: write` permission where it's needed on the individual jobs?

I think right now we allow all actions to run with `actions: write` level permissions, I'm not sure if we're not accidentally elevating actions that only need `read` level to higher permissions than they used to have?

> You can use `permissions` either as a top-level key, to apply to all jobs in the workflow, or within specific jobs. When you add the `permissions` key within a specific job, all actions and run commands within that job that use the `GITHUB_TOKEN` gain the access rights you specify.

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions

Belco90: It will be needed in all jobs in the workflow, that's why I put it as a top-level key.

HonkingGoose: Okay!

HonkingGoose (approved):

Belco90:
Btw you can see the cancel workflow action working in #258 since I trigger 2 workflows almost at the same time: when pushing changes and when updating my branch with `main`.

imagen

HonkingGoose:
Cool! So we know it works!

PR #514, opened 2021-11-02
Add custom 404 page

### LINKED ISSUE #258: Custom 404 page

Belco90:
## Changes

- add custom 404 page
- merge 2 links from the header into a single one (a11y improvement)
- improve SEO title

I've made some changes to the original design, so the banner image is not so big now. This way, the main message about the page wasn't found is readable without scrolling and everything fits better:

CleanShot 2021-11-02 at 01 09 15@2x

sizzy-photo-studio-2021-11-02T001018 835Z

## Context

Closes #258


src/pages/404.tsx
```diff
@@ -0,0 +1,61 @@
...
+              This page could not be found.
+            </Heading>
+            <Text as="p" color="coolGray.700" fontSize="2xl" fontWeight="black">
+              Octoclairvoyant cannot divine the page you wanted.
+            </Text>
+          </Stack>
+        </VStack>
+        <VStack mt={12} spacing={5}>
+          <NextLink href="/comparator">
+            <CTAButton>Go to comparator</CTAButton>
```
HonkingGoose, line 49:
```suggestion
            <CTAButton>Go to repository search</CTAButton>
```

I think "Go to repository search" is a better description than "Go to comparator". If you hit the 404 page, something went wrong, and we want people to recover quickly themselves.

Maybe using "Search" in the button is clearer on the action they can take after a user clicks the button?

If we put the `Go to repository search` into the button, it does break down on iPhone SE (1st gen) or smaller devices, where the button becomes too big too fit on screen comfortably.

I'm not sure how many viewports we want to define, each one we define is maintenance work.... So maybe we can just go with your text after all, and rely on `less text = smaller button`... 😄

src/pages/404.tsx
```diff
@@ -0,0 +1,61 @@
...
+import Link from '~/components/Link'
+
+const Custom404 = () => (
+  <Layout>
+    <NextSeo title="404 - Not Found" />
+    <Box pb={{ base: 8, lg: 16 }} align="center">
+      <FluidContainer>
+        <VStack px="10">
+          <Image
+            src={mascotLogo}
```
HonkingGoose, line 19: On Figma I used a placeholder image without the "Octoclairvoyant" text baked in. I think we can rely on the `h1` in the nav-bar to make clear that people are on the "Octoclairvoyant website".

The current image is too small/low resolution to use, but maybe you can make a new logo without the "Octoclairvoyant" text baked in?

Here's the image I used on Figma:

https://github.com/octoclairvoyant/octoclairvoyant-webapp/blob/main/public/mascot-icon.png

src/pages/404.tsx
```diff
@@ -0,0 +1,61 @@
...
+import mascotLogo from '@app-public/mascot-logo.png'
+import { CTAButton } from '~/components/CTAButton'
+import FluidContainer from '~/components/FluidContainer'
+import Layout from '~/components/Layout'
+import Link from '~/components/Link'
+
+const Custom404 = () => (
+  <Layout>
+    <NextSeo title="404 - Not Found" />
+    <Box pb={{ base: 8, lg: 16 }} align="center">
```
HonkingGoose, line 15: On the full-width page design on Figma, the mascot picture is 600x600px, to inject a bit of playfullness into the page.

I'm okay with scaling down the image on smaller viewports, but if we have room I'd like the image to take up more space.

We don't have many images anyways, so I'd like to show off those we do have.

What do you think?

src/next-seo.config.ts
```diff
@@ -3,8 +3,9 @@ import { DefaultSeoProps } from 'next-seo'
 import { BRIEF_DESCRIPTION, FULL_DESCRIPTION, SITE_TITLE } from '~/common'
 
 const DefaultSEO: DefaultSeoProps = {
-  titleTemplate: `%s | ${SITE_TITLE}: ${BRIEF_DESCRIPTION}`,
+  titleTemplate: `%s | ${SITE_TITLE}`,
   description: `🔮 ${FULL_DESCRIPTION}`,
+  defaultTitle: `${SITE_TITLE}: ${BRIEF_DESCRIPTION}`,
```
HonkingGoose, line 8: We need to update the Cypress homepage spec test, as it tests for the page title in there somewhere. 😄

HonkingGoose (commented):
Some initial comments, still reviewing the rest of the changes.

src/pages/404.tsx
```diff
@@ -0,0 +1,61 @@
...
+            </Text>
+          </Stack>
+        </VStack>
+        <VStack mt={12} spacing={5}>
+          <NextLink href="/comparator">
+            <CTAButton>Go to comparator</CTAButton>
+          </NextLink>
+
+          <NextLink href="/">
+            <Link fontWeight="black">Or go to homepage</Link>
```
HonkingGoose, line 53: For some reason I cannot use the keyboard with <kbd>tab</kbd> to focus on the `Or go to homepage` text.

Tabbing sequence:

1. `h1` in navbar
1. Developer `Toggle Dark` button
1. Big fat button `Go to comparator`
1. Skips over `Or go to homepage` text completely
1. Footer `Mario`
1. Footer `GitHub`
1. Vercel badge

src/next-seo.config.ts
Belco90: Yep, and add some tests for Custom 404 page.

src/pages/404.tsx
Belco90: Not really sure. I just realized your designs had the "Go to repository search" text, but it was natural for me to put "Go to comparator". I think we should refer to the page itself, which is the comparator. I find "repository search" quite confusing: it could be the comparator page starting by searching a repository, or another page to search repositories? I'd stick to "comparator", to be honest.

HonkingGoose: The idea was that people who land on the 404 might not even know what a comparator is, so that's why I thought maybe we could use `Go to repository search`, as that might be clearer for people new to Octoclairoyant.

The button is nice and big, so it'll be the first thing people will hit anyways, so maybe the text is less important.

It's also important to use a single word for a single concept, so let's keep `comparator` then. 😉

src/pages/404.tsx
Belco90: True, didn't realize about the image without text. Indeed, the current image icon is too small, but I have the original one and I can replace it to get a bigger one.

src/pages/404.tsx
Belco90: I mentioned the reason for this in the PR description:

> I've made some changes to the original design, so the banner image is not so big now. This way, the main message about the page wasn't found is readable without scrolling and everything fits better

I'd like to avoid opening a 404 error page just with a huge logo then having to scroll to understand what happened, it's quite confusing.

HonkingGoose: Oh okay, let's keep the focus on the error text instead of on the image then. I think you're probably right that people want to know what went wrong first, instead of spending too much time looking at a picture.

src/pages/404.tsx
Belco90: I don't know why this happens from time to time. It's the same issue you mentioned a few days ago about not being able to focus some links in the feature section of the homepage. I was able to do so on my other laptop, so there must be something preventing us to do so. I'll investigate it.

Belco90:
I can't find the way to test the custom 404 page with Cypress, I always get this:

CleanShot 2021-11-02 at 21 25 10@2x

I'll investigate more about it. Otherwise, it might be the perfect case to add a jest + Testing Library test.

Belco90:
I might need to check it by asserting the request body like here: https://www.cypress.io/blog/2019/09/04/guest-post-the-concept-of-monitoring-tests/#what-else-monitoring-tests-could-check

If that's the case, I definitely will test this with Testing Library instead.

package.json
```diff
@@ -17,8 +17,9 @@
     "format:check": "yarn prettier-base --check",
     "cypress": "cypress open",
     "cypress:headless": "cypress run",
-    "e2e": "start-server-and-test 'yarn dev' http://localhost:3000 'yarn cypress'",
-    "e2e:headless": "start-server-and-test 'yarn serve' http://localhost:3000 'yarn cypress:headless'",
+    "e2e": "start-server-and-test 'yarn build && yarn serve' http://localhost:3000 'yarn cypress'",
```
Belco90, line 20: Next.js v12 with SWC now builds the production bundle so fast that I've switched the default `e2e` command to build and serve the prod bundle rather than running the local dev server (still available under `e2e:dev` tho). This way, when navigating pages the E2E tests don't need to wait for building the page on-fly (causing delays in assertions etc).

HonkingGoose: Cool that means we could reduce the wait statements on Cypress as well, because I used that wait partially to account for the dev build when hitting the comparator page. We still need a reasonable wait for the GitHub API response though.

HonkingGoose:
> I can't find the way to test the custom 404 page with Cypress, I always get this:
> 
> CleanShot 2021-11-02 at 21 25 10@2x
> 
> I'll investigate more about it. Otherwise, it might be the perfect case to add a jest + Testing Library test.

Hmm, it looks like Cypress goes: "Received 404 status code, abort test" instead of going "404, follow redirect, check page content".

It says in the error message that you can ignore status codes, but that sounds like a bad idea, because for existing pages we _want_ the test to fail on 404 status code... 😄

Belco90:
> Hmm, it looks like Cypress goes: "Received 404 status code, abort test" instead of going "404, follow redirect, check page content".
> 
> It says in the error message that you can ignore status codes, but that sounds like a bad idea, because for existing pages we _want_ the test to fail on 404 status code... 😄

I see! It would be interesting to ignore the 404 code in this very specific test for testing purposes tho. I'll check that possibility.

src/pages/404.tsx
Belco90: Oh wow, the E2E test caught this too! I wasn't passing the href from the Next link to the Chakra link, ending in a wrong link from an a11y point of view. It should be fixed now!

cypress/integration/404.spec.ts
```diff
@@ -0,0 +1,23 @@
+it('should display custom 404 page when not found', () => {
+  cy.visit('/not-existing-page', { failOnStatusCode: false })
```
Belco90, line 2: I've set the `failOnStatusCode` here so I can test the custom 404 only in this very test.

HonkingGoose: Cool!

cypress/integration/404.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  cy.findByRole('heading', {
+    name: 'This page could not be found.',
+  }).should('exist')
+
+  cy.findByText('Octoclairvoyant cannot divine the page you wanted.').should(
+    'exist'
+  )
+
+  cy.findByRole('button', { name: 'Go to comparator' }).click()
+  cy.url().should('equal', `${Cypress.config().baseUrl}/comparator`)
```
HonkingGoose, line 15: I'm seeing: `${Cypress.config().baseURL}/comparator`in the test. 
In my tests I used another way to define the URL.

Is your way the best way to define the URL?

cypress/integration/404.spec.ts
```diff
@@ -0,0 +1,23 @@
+it('should display custom 404 page when not found', () => {
+  cy.visit('/not-existing-page', { failOnStatusCode: false })
+
+  cy.title().should('equal', '404 - Not Found | Octoclairvoyant')
+
+  cy.findByRole('heading', {
+    name: 'This page could not be found.',
+  }).should('exist')
+
```
HonkingGoose, line 9: I'd like to have a test to check if the correct image is visible on the 404 page here. 😉

HonkingGoose (changes requested):
Some initial test changes, I'll do a visual comparison with the Figma design later.

cypress/integration/404.spec.ts
Belco90: This way is better, so it always takes the `baseUrl` defined in `cypress.json`. Otherwise, if we hardcode the base URL and it's changed someday, all tests would start failing. I changed your test to use this one btw.

cypress/integration/404.spec.ts
Belco90: Sure, I'll add that assertion later.

cypress/integration/home.spec.ts
```diff
@@ -19,11 +19,9 @@ it('should should display corresponding information', () => {
...
     .should('include', 'mascot-logo')
 
-  cy.findByAltText(/octoclairvoyant reading a crystal ball/i).should(
-    'have.attr',
-    'src',
-    '/_next/image?url=%2F_next%2Fstatic%2Fmedia%2Fmascot-logo.8655d29c.png&w=1200&q=100'
-  )
+  cy.findByAltText(/octoclairvoyant reading a crystal ball/i)
+    .should('have.attr', 'src')
+    .and('contain', 'mascot-logo')
```
Belco90, line 24: I've updated this assertion too so we don't check the image src generated by Next.js but just the interesting bit.

HonkingGoose: That's way better!

.github/workflows/ci.yml
```diff
@@ -116,9 +116,3 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
           CYPRESS_PROJECT_ID: ${{ secrets.CYPRESS_PROJECT_ID }}
-          # - if the event is push, the title will be undefined
-          # and Cypress will get the commit message from Git information
-          # - if the event is pull_request, then we set the commit
-          # message to the pull request title
-          # https://github.com/cypress-io/github-action/issues/124#issuecomment-653180260
-          COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title }}
```
Belco90, line 124: I realized we should be getting some Cypress comments within GitHub PRs with E2E results. It turns out [this property is fixing such comments](https://github.com/cypress-io/github-action/issues/124#issuecomment-716584972), so I'm removing it since 1) we don't need it and 2) will fix the Cy comments.

Belco90: However, we get this ugly title for the run in Cy Dashboard instead:

CleanShot 2021-11-03 at 16 08 11@2x

Which option you prefer @HonkingGoose ?

Belco90: Reverting for now since the commit info message seems more useful, and we can still access the Dashboard results from Actions within the PR.

cypress/integration/404.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  cy.findByText('Octoclairvoyant cannot divine the page you wanted.').should(
+    'exist'
+  )
+
+  cy.findByAltText(/octoclairvoyant reading a crystal ball/i)
```
HonkingGoose, line 14: I think we could improve our `alt` text, consider that people who do not see the image, do not know what an "Octoclairvoyant" is.

How about this `alt` text: "A purple octopus reading a crystal ball"?

https://webaim.org/techniques/alttext/

src/components/Link.tsx
```diff
@@ -1,5 +1,10 @@
 import { Link as ChakraLink, LinkProps } from '@chakra-ui/react'
+import { forwardRef } from 'react'
 
-const Link = (props: LinkProps) => <ChakraLink color="primary.500" {...props} />
+const Link = forwardRef<HTMLAnchorElement, LinkProps>((props, ref) => (
```
HonkingGoose, line 4: I'm still having problems using <kbd>tab</kbd> with Firefox `94.0`.
I can now only focus on the big "Go to comparator" button with the keyboard, not on the "Or go to homepage" text, or on the h1 navbar, or even the links in the footer.

So something is still broken here. 😕

cypress/integration/home.spec.ts
```diff
@@ -19,11 +19,9 @@ it('should should display corresponding information', () => {
     .should('have.attr', 'src')
     .should('include', 'mascot-logo')
 
-  cy.findByAltText(/octoclairvoyant reading a crystal ball/i).should(
-    'have.attr',
-    'src',
-    '/_next/image?url=%2F_next%2Fstatic%2Fmedia%2Fmascot-logo.8655d29c.png&w=1200&q=100'
-  )
+  cy.findByAltText(/octoclairvoyant reading a crystal ball/i)
```
HonkingGoose, line 22: We should probably improve this alt text as well while we're at it.

HonkingGoose (changes requested):
Found some more accessibility problems unfortunately.

cypress/integration/404.spec.ts
Belco90: Happy to improve that, although it's not an octopus but an octocat 😅

src/components/Link.tsx
Belco90: I don't see more issues navigating with the keyboard. This is on firefox 94:

https://user-images.githubusercontent.com/2677072/140300984-68173712-23fa-4081-baa4-1d564d35ade0.mp4

This might be a problem on your end. Make sure you have enabled the browser keyboard navigation if you are using macOS: https://www.a11yproject.com/posts/2017-12-29-macos-browser-keyboard-navigation/

I already had this + always visible scrollbar in order to test websites properly. I just checked that disabling that option I can't navigate to the links either, so it must be definitely that.

cypress/integration/404.spec.ts
HonkingGoose: Do you mean it's like the GitHub mascot then? They call that an octocat as well? I'm missing the cat like ears then on our mascot???

src/components/Link.tsx
HonkingGoose: > This might be a problem on your end. Make sure you have enabled the browser keyboard navigation if you are using macOS: https://www.a11yproject.com/posts/2017-12-29-macos-browser-keyboard-navigation/

Problem was on my end. 🙈 Fixed now by following the instructions in the link! 🥳

Belco90: This has been a huge TIL for me by the way! I didn't know that option could affect the keyboard navigation in websites, I thought it was meant to be only for macOS dialogs (which is the reason I enabled the setting).

cypress/integration/404.spec.ts
Belco90: Yeah like the GitHub mascot, but it was complicated to explain it to the designer that made the logo 😂 

I'll keep it simple and just refer to it as "octopus" tho. I'll push the change in a bit.

src/pages/404.tsx
```diff
@@ -0,0 +1,61 @@
...
+          <Stack
+            shouldWrapChildren
+            alignItems="center"
+            spacing="2"
+            flexGrow={1}
+            flexBasis={0}
+          >
+            <Heading
+              as="h2"
+              color="gray.900"
```
HonkingGoose, line 35:
```suggestion
              color="primaryTextLightmode"
```

src/pages/404.tsx
```diff
@@ -0,0 +1,61 @@
...
+              fontSize="4xl"
+              fontWeight="black"
+              letterSpacing="tight"
+            >
+              This page could not be found.
+            </Heading>
+            <Text as="p" color="coolGray.700" fontSize="2xl" fontWeight="black">
```
HonkingGoose, line 42:
```suggestion
            <Text as="p" color="secondaryTextLightmode" fontSize="2xl" fontWeight="black">
```

HonkingGoose (changes requested):
We should use our pre-defined colors for UI text. 😉 

I defined 6 shades for us:
https://github.com/octoclairvoyant/octoclairvoyant-webapp/blob/330c06024b04d535ad15839faebc46d387c255f0/src/customTheme.ts#L115-L120

We do run the risk of "painting ourselves into a corner here" as we need a way to switch between light and dark mode UI text colors programatically. But I think you already had ideas on that?

Belco90:
> We do run the risk of "painting ourselves into a corner here" as we need a way to switch between light and dark mode UI text colors programatically. But I think you already had ideas on that?

Yes, I have a bare idea of how to do it. I expect to figure out more about it on #515, but basically, the idea will be that just using the token `"primaryText"` gets the corresponding color for either light or dark mode (same for secondary and tertiary variants).

src/pages/404.tsx
Belco90: Forgot about this token shortcut!

HonkingGoose:
> > We do run the risk of "painting ourselves into a corner here" as we need a way to switch between light and dark mode UI text colors programatically. But I think you already had ideas on that?
> 
> Yes, I have a bare idea of how to do it. I expect to figure out more about it on #515, but basically, the idea will be that just using the token `"primaryText"` gets the corresponding color for either light or dark mode (same for secondary and tertiary variants).

I like this idea a lot! Easy to write, easy to think about, easy to maintain. 🥳

HonkingGoose (approved):
I think we're good to go on this!

PR #518, opened 2021-11-02
Add tests section to PR template

HonkingGoose:
## Changes

- Add section asking about tests to PR template

## Context

The goal is to make sure we:

- Do not introduce new stuff without tests
- Focus on testing existing stuff before updating/refactoring existing features

To reach this goal I've added a reminder to our PR template, that explains that we do not want features without tests, and ask how the PR author has tested their work.


.github/pull_request_template.md
```diff
@@ -8,3 +8,13 @@
...
+
+## Tests
+
+<!-- We do not accept new features without corresponding automated tests. -->
+
+I've checked my work by:
+
+- [ ] Adding new test or adjusting existing tests
+- [ ] Testing manually in the browser
+- [ ] Looking at the code only
```
Belco90, line 20: What does this mean?

HonkingGoose: The idea here is that people can put a `X` in the box, to say how they've tested their work.
The goal is for us to get a at-a-glance idea of how well tested the change is.

### Example: fixing typo in markdown file

Say I'd fixed a simple typo in the README, then I'd say:

> - [X] Looking at the code only

### Example: create new page

But if I made a new 404 page, I'd need to have:

> - [X] Adding new test or adjusting existing tests

### Example: new feature without tests

If I made a new feature and said:

> - [X] Looking at the code only

Then we could say: "Go write some tests".

Belco90: Oh, I see. I guess you see this as a single choice (radio buttons), but those are checkbox so I always think about this as non-excluding options.

I'd leave it to just `Adding new test or adjusting existing tests` and `Testing manually` for simplicity? So testing manually covers manual checks in the browser or checking something in the code. Then if there is a new feature added and the first checkbox is not check, we can say "hey this needs tests".

HonkingGoose, line 20:
```suggestion
- [ ] Adding new test or adjusting existing tests
- [ ] Testing manually
```

HonkingGoose: Let's simplify then. 😄

.github/pull_request_template.md
```diff
@@ -8,3 +8,12 @@
...
 Fixes #1
 -->
+
+## Tests
+
+<!-- We do not accept new features without corresponding automated tests. -->
+
+I've checked my work by:
+
+- [ ] Adding new test or adjusting existing tests
```
HonkingGoose, line 18:
```suggestion
- [ ] Adding new tests or adjusting existing tests
```

Belco90 (approved):
LGTM!

PR #542, opened 2021-11-09
remove google analytics

Belco90:
## Changes

Removing Google Analytics since we are not using it at all, but makes the initial load of the app worse.

If at some point we need analytics of the app, we should explore some alternatives to gtag.

## Context

n/a

## Tests

I've checked my work by:

- [X] Testing manually


.gitignore
```diff
@@ -19,3 +19,4 @@ cypress.env.json
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+/tsconfig.tsbuildinfo
```
HonkingGoose, line 22: There's a new line in the `.gitignore` is that intentional?

I think this has to do with TypeScript incremental builds?

HonkingGoose (commented):
Looks good, but is that change in the `.gitignore` intentional?

.gitignore
Belco90: Yeah, it was on purpose. [This file shouldn't be included in VCS](https://betterstack.dev/blog/lerna-typescript-monorepo/).

HonkingGoose: Okay! 😄

HonkingGoose (approved):

PR #545, opened 2021-11-10
Move from yarn v1 to npm v8

### LINKED ISSUE #544: Migrate from Yarn 1 to npm 8

We need to migrate from Yarn v1 into something else.

[We could migrate to Yarn v3](https://yarnpkg.com/getting-started/migration), although I'm not a huge fan of v2/3 changes, to be honest. [You also need npm itself to run some commands now](https://yarnpkg.com/getting-started/migration#cli-commands).

npm is a great option now, so we could migrate to npm v8 since 1) it's needed by Yarn v3 for some commands, 2) it's the default package manager you get when installing node, and 3) seems a less-risky option.

What do you think @HonkingGoose?

[The yarn 1.x line is frozen](https://github.com/yarnpkg/yarn,) so it would be great to migrate to something else.

Yes

_No response_

Belco90:
## Changes

- move the package manager from yarn v1 to npm v8

## Context

Closes #544 

## Tests

I've checked my work by:

- [X] Testing manually


Belco90:
This is ready now to get reviewed, sorry for the noise while fixing the lock file 😅

.github/workflows/ci.yml
```diff
@@ -8,6 +8,9 @@ permissions:
   contents: read
   actions: write
 
+env:
+  NODE_VERSION: '14'
```
Belco90, line 12: Just extracting it to an env var so we only need to change it in a single place in the future.

.npmrc
```diff
@@ -0,0 +1 @@
+save-exact=true
```
Belco90, line 1: The only option we had in `.yarnrc` migrated to npm.

package.json
```diff
@@ -1,28 +1,30 @@
 {
-  "name": "octoclairvoyant",
+  "name": "octoclairvoyant-webapp",
   "version": "1.0.0",
   "private": true,
+  "engines": {
+    "npm": "^8.0.0"
```
Belco90, line 6: This is just a hint, it won't actually throw an error if the package manager + version is not matched. When we upgrade to Node v16, it might be useful to set up [Corepack](https://nodejs.org/dist/latest-v16.x/docs/api/corepack.html).

package.json
```diff
@@ -1,28 +1,30 @@
...
   "scripts": {
-    "start": "yarn dev",
+    "start": "npm run dev",
     "dev": "next",
     "build": "next build",
     "serve": "next start",
-    "smoketest": "yarn type-check && yarn lint && yarn format:check && yarn build",
+    "smoketest": "npm run type-check && npm run lint && npm run format:check && npm run build",
     "type-check": "tsc --noEmit",
-    "analyze": "ANALYZE=true yarn build",
```
Belco90, line 12: This one isn't valid anymore since the `@next/bundle-analyzer` was removed.

.github/workflows/ci.yml
HonkingGoose: That's a lot better if we need to change it later! Good catch!

.prettierignore
```diff
@@ -1,3 +1,4 @@
 yarn.lock
```
HonkingGoose, line 1:
```suggestion
```

We can remove `yarn.lock` from the `.prettierignore` now. 😉

CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ It's even better if you know a bit of React and the basics of TypeScript.
 
 - Git
 - Latest version of Node.js 14
-- Yarn v1 package manager
+- npm v8 package manager (should be provided by Node.js)
```
HonkingGoose, line 20:
```suggestion
- npm v8 package manager (bundled with Node.js)
```

CONTRIBUTING.md
```diff
@@ -27,18 +27,16 @@ The [git-scm.com](https://git-scm.com/) site has links to download Git for your
 
 ### Where to get Node.js
 
+You can install Node.js + npm using a [Node.js version manager](https://docs.npmjs.com/cli/v7/configuring-npm/install#using-a-node-version-manager-to-install-nodejs-and-npm) (recommended).
+
 If you use Linux, use your operating system's package manager (`dnf`, `apt` or similar) to install the latest version of Node 14.
 If you use Ubuntu, you can also get a snap of the Node 14 program.
 
 If you use Windows or macOS, you can grab binaries from the [Node.js offical website](https://nodejs.org/).
```
HonkingGoose, line 35:
```suggestion
```

Let's drop these lines and leave only the Node version manager route in the docs.

Quote from the npm docs:

> **We strongly recommend using a Node version manager to install Node.js and npm.**

package.json
HonkingGoose: The hinting might be helping Renovate bot to find the right package manager to run, so let's leave it in.

That corepack thing seems nice:

> 
> This feature simplifies two core workflows:
> 
> - It eases new contributor onboarding, since they won't have to follow system-specific installation processes anymore just to have the package manager you want them to.
> 
> - It allows you to ensure that everyone in your team will use exactly the package manager version you intend them to, without them having to manually synchronize it each time you need to make an update.

It seems `corepack` is still experimental, so we should probably wait until the docs say it's stable. 😄

HonkingGoose (changes requested):

HonkingGoose (commented):
I tried running the branch in Gitpod, during the `npm install` step, I get this warning:

```bash
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

The Gitpod container uses Node 14.18.1 and npm 6.14.15 by default.
We probably need to change the `.gitpod.Dockerfile` to make changes so the Gitpod container uses the correct node and npm?

.gitpod.Dockerfile
```diff
@@ -66,6 +66,9 @@ RUN sudo apt-get update \
 # Source of original file https://github.com/gitpod-io/website/blob/main/src/routes/docs/languages/javascript.md
 
 # Use nvm to setup the version of node we want to use.
-RUN bash -c ". .nvm/nvm.sh     && nvm install 14     && nvm use 14     && nvm alias default 14"
+RUN bash -c ". .nvm/nvm.sh \
+    && nvm install \
+    && nvm use \
+    && npm install -g npm@8"
```
Belco90, line 73: Can you try if it works ok with this updated command?

Specifying node version shouldn't be necessary since that's already done in `.nvmrc`, and upgrading npm itself should be as simple as the last command.

HonkingGoose: It's not working at all now:

```bash
No .nvmrc file found

Node Version Manager (v0.39.0)

Note: <version> refers to any version-like string nvm understands. This includes:
  - full or partial version numbers, starting with an optional "v" (0.10, v0.1.2, v1)
  - default (built-in) aliases: node, stable, unstable, iojs, system
  - custom aliases you define with `nvm alias foo`
```

```bash
Error: build failed: cannot build base image: The command '/bin/sh -c bash -c ". .nvm/nvm.sh     && nvm install     && nvm use     && npm install -g npm@8"' returned a non-zero code: 127
```

I think there's a problem with finding the `.nvmrc` file?

Belco90: Ah yeah, I might need to copy that file into the image. I'll give it a try later.

.gitpod.Dockerfile
```diff
@@ -66,6 +66,10 @@ RUN sudo apt-get update \
...
 # Use nvm to setup the version of node we want to use.
-RUN bash -c ". .nvm/nvm.sh     && nvm install 14     && nvm use 14     && nvm alias default 14"
+COPY .nvmrc ./
+RUN bash -c ". .nvm/nvm.sh \
+    && nvm install \
+    && nvm use"
+RUN echo "nvm use &>/dev/null" >> ~/.bashrc.d/51-nvm-fix
 
-RUN echo "nvm use default &>/dev/null" >> ~/.bashrc.d/51-nvm-fix
+RUN npm install -g npm@8
```
HonkingGoose, line 75: Now the build works again, but I still get the shrinkwrap warning:

```bash
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

Running `node -v` and `npm -v` give these outputs:

```bash
gitpod /workspace/octoclairvoyant-webapp $ node -v
v14.18.1
gitpod /workspace/octoclairvoyant-webapp $ npm -v
6.14.15
```

So Gitpod still doesn't use the right version of npm when it starts up.

Belco90: Too soon, try now 😉

HonkingGoose: That fixed it! 🥳

HonkingGoose (approved):

PR #575, opened 2021-11-20
Add more footer tests

HonkingGoose:
## Changes

- Check page shows link to Mario's homepage
- Check page shows link to Octoclairvoyant's GitHub repository
- Check page shows Vercel badge

## Context

This PR adds some more tests to check if we're actually displaying the wanted items in our footer. Right now it uses find by role, but it's not limited to checking the `<footer>` section, so it's really checking if the page shows the item, not if the footer specifically contains the items.

I need some help to get the Vercel badge test working properly. I cannot get it to accurately find the `src` to test that it uses the correct SVG image. I also cannot test the `href` that comes with the Vercel badge.

## Tests

I've checked my work by:

- [x] Adding new tests or adjusting existing tests
- [ ] Testing manually


cypress/integration/home.spec.ts
```diff
@@ -26,8 +26,29 @@ it('should display corresponding information', () => {
...
+  }).should(
+    'have.attr',
+    'href',
+    'https://github.com/octoclairvoyant/octoclairvoyant-webapp'
+  )
+
+  cy.findByRole('img', { name: /powered by vercel logo/i })
+  // Find way to make this query below work
+  // .should('have.attr', 'src')
+  // .should('include', 'powered-by-vercel')
```
Belco90, line 47: Since it's an svg element, it might need a different way of checking the attributes.

Belco90: I can't find any reason why we are getting the base64 image instead of the URL here (might be related to how Next.js + Vercel works with SVG locally vs deploying the real app). I'd just skip the `src` check on this logo, and check it has the expected url.

Belco90:
Is this PR ready for review? I thought there was some assertion still pending.

cypress/integration/home.spec.ts
```diff
@@ -26,6 +26,24 @@ it('should display corresponding information', () => {
...
+    name: /octoclairvoyant repository on github/i,
+  }).should(
+    'have.attr',
+    'href',
+    'https://github.com/octoclairvoyant/octoclairvoyant-webapp'
+  )
+
+  cy.findByRole('img', { name: /powered by vercel logo/i })
+
+  // TODO: Add test that checks if the Vercel img has the correct href, should be: https://vercel.com/?utm_source=octoclairvoyant-team&utm_campaign=oss
```
HonkingGoose, line 46: The thing that's missing here is the assertion that checks if the Vercel `` has the correct `href`. I still don't know how to check for this, so that's whats holding up this PR from getting merged.

Belco90: Aah, I see. The image itself doesn't have any `href` attribute, so we need to query the link for that. Since the image is inside the link, the best strategy is to query the link and assert whatever we want, and then query the image inside the link element obtained before and assert whatever we want for the image. If we remove the previous assert to find the Vercel logo, it can be replaced with this:

```js
// we first query the link
cy.findByRole('link', { name: /powered by vercel/i })
    // and check it has the proper href
    .should(
      'have.attr',
      'href',
      'https://vercel.com/?utm_source=octoclairvoyant-team&utm_campaign=oss'
    )
    // now, since `should` returns a Chainable, we can chain it with `within`, which allows us to query
    // for something inside the previous element
    .within(() => {
      // the image is gonna be searched within the link, so we can make sure the logo is part of the link!
      cy.findByRole('img', { name: /powered by vercel logo/i })
    })
```

HonkingGoose: Ah that works! I didn't know about the `.within()` thing! 😄 

Thanks for the comments, I've left them in the production code, as I will probably need to refer to this later on for other tests... 🙈 😄

Belco90 (approved):

PR #643, opened 2022-01-03
Minor improvements

Belco90:
## Changes

- remove unnecessary `@types/nprogress` dependency: this was some leftover when we were using `nprogress`
- enable some new typescript-eslint rules:
  - `@typescript-eslint/no-floating-promises`
  - `@typescript-eslint/consistent-type-exports`
  - `@typescript-eslint/consistent-type-imports`
  - `@typescript-eslint/no-explicit-any`
- remove duplicated ESLint plugins. All plugins already installed by `eslint-next-config` have been removed
- remove duplicated ESLint config. All config already setup by `eslint-next-config` has been removed

## Context

Just some minor improvements

## Tests

I've checked my work by:

- [X] Testing manually


.eslintrc.js
```diff
@@ -1,20 +1,13 @@
 module.exports = {
-  root: true,
-  parser: '@typescript-eslint/parser',
   extends: [
     'eslint:recommended',
+    'next/core-web-vitals',
```
HonkingGoose, line 4: This `next/core-web-vitals` preset seems to replace the old `next` preset, and comes with some stricter rules, I think?

Quote from Next.js docs that cover ESLint setup [^docs]:

> **Strict**: Includes Next.js' base ESLint configuration along with a stricter Core Web Vitals rule-set. This is the recommended configuration for developers setting up ESLint for the first time.
> 
> ```json
> {
>   "extends": "next/core-web-vitals"
> }
> ```
> 

[^docs]: https://nextjs.org/docs/basic-features/eslint#eslint-plugin

next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
```
HonkingGoose, line 4: There's a comment from Next.js here that says not to edit the file...

Are you sure that you want to make this change?

package.json
```diff
@@ -13,7 +13,7 @@
     "smoketest": "npm run type-check && npm run lint && npm run format:check && npm run build",
     "type-check": "tsc --noEmit",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx --max-warnings 0 --report-unused-disable-directives",
-    "lint:fix": "npm run lint --fix",
+    "lint:fix": "npm run lint -- --fix",
```
HonkingGoose, line 16: I've seen this double `--` thing multiple times now, and I'm still not sure what it's good for... 😄 

I'm going to assume you tested that this works properly. 😉

package.json
```diff
@@ -60,24 +60,16 @@
     "@testing-library/cypress": "8.0.2",
     "@types/lodash": "4.14.178",
     "@types/node": "14.18.3",
-    "@types/nprogress": "0.2.0",
```
HonkingGoose, line 63: Goodby old depenency! 😄

.eslintrc.js
Belco90: Yes, basically this preset enables the basic `next` one plus some other rules for core web vitals. This is a more extended preset.

next-env.d.ts
Belco90: Oh sorry, I didn't modify this myself. After removing the `@types/nprogress` one, that line got removed by next itself.

HonkingGoose: That's fine then, if the program did it. 😉

package.json
Belco90: The double `--` is a bit difficult to explain, but I'll do my best:

With the previous command `npm run lint --fix` we were passing the fix option to the npm script itself. But what we want to do is to pass the fix option to the command executed under the hood, ESLint in this case. To do so, you need to indicate that with a `--` before whatever you want to pass to such command. Now, we are passing the fix option properly to ESLint with this trick!

HonkingGoose: Ah that's a lot clearer now! Thank you! 😄

HonkingGoose (approved):
I think this is good to go! 😄

PR #653, opened 2022-01-09
Remove repository's name capitalization

Belco90:
## Changes

- Avoid capitalizing repository name in the main header after selecting some repo

## Context

There was no ticket for this. It was a bit too much to have `title` dependency for this, so I removed it and leave the original repo name.

## Tests

I've checked my work by:

- [ ] Testing manually

### DIFF
--- a/additional.d.ts
+++ b/additional.d.ts
@@ -1,6 +1,1 @@
-declare module 'title' {
-  export default function title(
-    str: string,
-    config: { special: Array<string> }
-  ): string
-}
+// declare here any additional module with no types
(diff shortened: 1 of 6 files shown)


HonkingGoose:
> There was no ticket for this. It was a bit too much to have `title` dependency for this, so I removed it and leave the original repo name.

If capitalizing is all the dependency does, it does seem a bit much to have that for this one feature of the dependency.

So we're just using the GitHub repository name now?

Belco90:
> So we're just using the GitHub repository name now?

That's right.

HonkingGoose (approved):

Issue #510, opened 2021-11-01
Unit/integration tests with Jest + Testing Libary


HonkingGoose:
### What feature would you like to have?

We probably should have some unit/integration tests which use Jest + Testing Library. Relying on Cypress for most tests will be too slow, and will not test the internals of the app.

### What problem will this feature solve?

Increased confidence in the app.

### Do you want to help make the feature?

Yes

Belco90:
Follow [this guide](https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler) when the time comes.

PR #659, opened 2022-01-10
Setup jest

Belco90:
## Changes

- Setup jest with [Next.js + rust compiler](https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler)
- Setup jest with TypeScript and ESLint
- Setup tests job in CI
- Add first test suite for `mapRepositoryToQueryParams` util

## Context

Relates to #510

This doesn't cover the whole scope of the ticket, but integrates jest in the codebase. To prove this, I added some tests for one of our own utils functions. In the following PRs I'll add more tests to cover all the utils we have (they are quite important and any stupid change to them could drastically change the app behavior or break it), and then setup React Testing Library plus some components tests.

## Tests

I've checked my work by:

- [ ] Testing manually


.eslintrc.js
```diff
@@ -42,7 +48,7 @@ module.exports = {
     {
       files: ['**/*.ts?(x)'],
       parserOptions: {
-        project: ['./tsconfig.json'],
+        project: ['./tsconfig.lint.json'],
```
Belco90, line 51: I had to generate different TS config files depending on their purpose to avoid clashing types between Cypress and Jest. You can read more about it here: https://docs.cypress.io/guides/tooling/typescript-support#Clashing-types-with-Jest

.github/workflows/ci.yml
```diff
@@ -48,8 +48,10 @@ jobs:
         run: npm ci
 
       # Act
-      - name: Check types
+      - name: Check codebase types
         run: npm run type-check
+      - name: Check Cypress types
+        run: npm run type-check:cypress
```
Belco90, line 54: I had to split the types validation into two commands, so now we need to run those 2 here.

src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,29 @@
+import type { Repository } from '../models'
```
Belco90, line 1: First tests in jest!

Belco90:
When this PR is about to get merged, I'll flag the new tests job required for PRs.

package.json
```diff
@@ -57,6 +61,7 @@
   "devDependencies": {
     "@chakra-ui/cli": "1.7.0",
     "@testing-library/cypress": "8.0.2",
```
HonkingGoose, line 63: The Next.js Jest testing guide says to install [^guide]:

- `jest` <- Installed
- `@testing-library/react` <- Not installed
- `@testing-library/jest-dom` <- Not installed

Do you want to setup the `@testing-libary` stuff later on?

[^guide]: https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler

src/__tests__/utils.test.ts
HonkingGoose: Hooray!

tsconfig.lint.json
```diff
@@ -0,0 +1,16 @@
+/* Avoid clashing types between Cypress and Jest for linting files
```
HonkingGoose, line 1: It's always annoying that you're not allowed to use comments in JSON officially...

If the comments are not causing us problems when running programs, I'd like to keep them in, so that we know what we did when we read this code later. 😄

HonkingGoose (commented):
Some initial comments.

package.json
Belco90: That's right, I'll setup React Testing Library in a future PR.

I'm just testing plain JS functions for now, but we need to transform TS tests to be run with Jest. Instead of configuring that manually, I used Next built-in config for Jest.

tsconfig.lint.json
Belco90: It's fine for TS config files since they are based in json5, which allows comments.

package.json
HonkingGoose: > That's right, I'll setup React Testing Library in a future PR.

Cool!

> Instead of configuring that manually, I used Next built-in config for Jest.

Yeah better to use built-in config than cobble something together yourself.

tsconfig.lint.json
HonkingGoose: Let's leave the comments in then! 😄

src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    expect(result).toEqual({ owner: 'foo', repo: 'bar' })
+  })
+
+  it('should map a partial repository', () => {
+    const result = mapRepositoryToQueryParams({
+      owner: {},
+      name: 'bar',
+    } as Repository)
+
+    expect(result).toEqual({ owner: '', repo: 'bar' })
```
HonkingGoose, line 27: The input data for the tests is similar, but not 100% the same. I'm wondering if we could simplify the setup steps by putting the data in a table?

Example from Jest docs:

```js
test.each`
  a    | b    | expected
  ${1} | ${1} | ${2}
  ${1} | ${2} | ${3}
  ${2} | ${1} | ${3}
`('returns $expected when $a is added $b', ({a, b, expected}) => {
  expect(a + b).toBe(expected);
});
```

> Use `test.each` if you keep duplicating the same test with different data. `test.each` allows you to write the test once and pass data in.

https://jestjs.io/docs/api#testeachtablename-fn-timeout

Belco90: Yes, we definitely could. I didn't want to put that effort into tests for this PR. I'll manage this in the next round of changes, where I'll add tests for the whole file.

HonkingGoose (approved):
I think this is good to go now. The stuff I mentioned can be done in a new PR later.

PR #660, opened 2022-01-11
Visual Testing tool with Percy

### LINKED ISSUE #511: Visual diff testing

Right now we need to test manually to see if the layout/visuals of the app change whenever we update our dependencies.

I think we could probably do some sort of "visual snapshot" testing?

Increased confidence updates do not break our layout/visuals.

Yes

_No response_

Belco90:
## Changes

- Setup Percy in the codebase and CI
- Add first snapshot with Percy

## Context

Closes #511

We have 5,000 screenshots per month with our current plan. Every `cy.percySnapshot` usage would count as 8 screenshot since we are getting screenshots for 2 widths and 4 browsers. I believe we have enough for now.

## Tests

I've checked my work by:

- [ ] Testing manually

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
           group: merge
           # tag will be either "push" or "pull_request"
           tag: ${{ github.event_name }}
-          command-prefix: 'percy exec -- npx'
+          command-prefix: 'percy exec -t 350 -- npx'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,7 @@ jobs:
       fail-fast: false
       matrix:
         # run 3 copies of the current job in parallel
+        # (remember to update PERCY_PARALLEL_TOTAL if this is changed)
         containers: [1, 2, 3]
 
     steps:
@@ -149,5 +150,6 @@ jobs:
 
           # Percy config
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+          PERCY_PARALLEL_TOTAL: 3 # remember to keep this synced with matrix.containers
           PERCY_TARGET_BRANCH: 'main'
           PERCY_BRANCH: ${{ github.head_ref }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,5 @@ jobs:
           # Percy config
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
           PERCY_PARALLEL_TOTAL: 3
-          PERCY_PARALLEL_NONCE: '${{ github.event_name }}-${{ github.sha }}'
           PERCY_TARGET_BRANCH: 'main'
           PERCY_BRANCH: ${{ github.head_ref }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,6 @@ jobs:
       fail-fast: false
       matrix:
         # run 3 copies of the current job in parallel
-        # (remember to update PERCY_PARALLEL_TOTAL if this is changed)
         containers: [1, 2, 3]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,5 @@ jobs:
 
           # Percy config
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
-          PERCY_PARALLEL_TOTAL: 3
           PERCY_TARGET_BRANCH: 'main'
           PERCY_BRANCH: ${{ github.head_ref }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,7 @@ jobs:
       fail-fast: false
       matrix:
         # run 3 copies of the current job in parallel
+        # (remember to update PERCY_PARALLEL_TOTAL if this is changed)
         containers: [1, 2, 3]
 
     steps:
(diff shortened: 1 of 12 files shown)


Belco90:
Here we have Percy letting us know about the first snapshot!
CleanShot 2022-01-11 at 16 55 48@2x

cypress/integration/404.spec.ts
```diff
@@ -23,5 +23,3 @@ it('should display custom 404 page when not found', () => {
   cy.findByRole('link', { name: 'Or go to homepage' }).click()
   cy.url().should('equal', `${Cypress.config().baseUrl}/`)
 })
-
-export {}
```
Belco90, line 27: This weird `export {}` is not needed anymore. Check the `cypress/tsconfig.json` changes for more info.

cypress/integration/comparator.spec.ts
```diff
@@ -94,8 +94,6 @@ it('should show expected results when using standard query string', () => {
   cy.findByRole('heading', { level: 2, name: /recommendations/i })
 
   cy.findByRole('heading', { level: 2, name: /chore/i })
-})
-
-// Keep the export {} braces here!
 
-export {}
+  cy.percySnapshot()
```
Belco90, line 98: Taking a snapshot of the comparator after all existing checks.

cypress/integration/home.spec.ts
```diff
@@ -55,6 +55,8 @@ it('should display corresponding information', () => {
       // the image is gonna be searched within the link, so we can make sure the logo is part of the link!
       cy.findByRole('img', { name: '' }).should('have.attr', 'src')
     })
+
+  cy.percySnapshot()
```
Belco90, line 59: Taking a snapshot of the landing page after all existing checks.

cypress/tsconfig.json
```diff
@@ -3,9 +3,10 @@
   "extends": "../tsconfig.json",
   "compilerOptions": {
     "noEmit": true,
+    "isolatedModules": false,
```
Belco90, line 6: This was the option requiring the `export {}` at the bottom of the Cypress tests. This option is useful in the main codebase (where each file represents a module), but not in Cypress tests, so I've disabled it here.

package.json
```diff
@@ -23,11 +23,13 @@
     "test:ci": "npm run test -- --ci",
     "cypress": "cypress open",
     "cypress:headless": "cypress run",
+    "cypress:percy": "percy exec -- cypress run",
```
Belco90, line 26: We shouldn't need to run Percy locally, but I added some commands just in case. Still, you need to prepend the env var with the API key to be able to run it locally.

package.json
```diff
@@ -23,11 +23,13 @@
...
     "e2e": "start-server-and-test 'npm run build && npm run serve' http://localhost:3000 'npm run cypress'",
     "e2e:dev": "start-server-and-test 'npm run dev' http://localhost:3000 'npm run cypress'",
     "e2e:headless": "start-server-and-test 'npm run build && npm run serve' http://localhost:3000 'npm run cypress:headless'",
+    "e2e:percy": "start-server-and-test 'npm run build && npm run serve' http://localhost:3000 'npm run cypress:percy'",
     "gen:theme-typings": "chakra-cli tokens src/customTheme.ts --strict-component-types",
-    "prepare": "husky install",
+    "prepare": "is-ci || husky install",
```
Belco90, line 32: Just a small improvement to husky so we don't install it on CI, which should speed up a bit the dependencies installation.

Belco90:
I need to investigate why the assets are missing in all browsers except Chrome 🤔

Belco90:
I added some allowed hosts so Percy can retrieve the assets, but it's not working.

I've contacted support, I hope they can help me to solve this.

next.config.mjs
```diff
@@ -7,9 +7,6 @@ import { withSentryConfig } from '@sentry/nextjs'
  **/
 const config = {
   swcMinify: true,
-  images: {
-    formats: ['image/avif', 'image/webp'],
-  },
```
Belco90, line 12: I figured out why images weren't loaded properly in browsers other than Chrome within Percy: because we were generating optimized images in AVIF format. I've disabled such optimization, so they are generated in Webp and it's working for all browsers in Percy.

I still need to figure out why Firefox doesn't load fonts properly, but that's a minor issue. We can go ahead now.

HonkingGoose: > I still need to figure out why Firefox doesn't load fonts properly, but that's a minor issue. We can go ahead now.

I use Firefox on my computer and the homepage looks like the Percy snapshots for the Chrome, Safari and Edge browsers.

It looks to me that this is a problem with running Firefox within Percy somehow?

.github/workflows/ci.yml
```diff
@@ -145,3 +147,10 @@ jobs:
           # message to the pull request title
           # https://github.com/cypress-io/github-action/issues/124#issuecomment-653180260
           COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title }}
+
+          # Percy config
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+          PERCY_PARALLEL_TOTAL: 3
+          PERCY_PARALLEL_NONCE: '${{ github.event_name }}-${{ github.sha }}'
```
HonkingGoose, line 154: The Percy docs say that `PERCY_PARALLEL_NONCE` does not need to be set in most cases.
Are we sure we need this property, maybe Percy can figure it out itself?

Quote from the docs: [^docs]

> `PERCY_PARALLEL_NONCE`: A unique identifier used to group parallel builds. This can be any string that is shared between parallel runners, such as the CI build number.
>
> -In most setups this is automatically discovered and does not need to be set. See Parallel test suites for more info.

[^docs]: https://docs.percy.io/docs/environment-variables#optional

next.config.mjs
Belco90: Yes, it's a problem within Percy. We can go ahead and merge this branch tho, then we will fix Firefox screenshots when we figure out the issue with support's help.

.github/workflows/ci.yml
Belco90: I didn't realize about this. Let me remove it to see if it works as expected.

next.config.mjs
HonkingGoose: Sounds good to me!

.github/workflows/ci.yml
Belco90: Everything keeps working as expected! Good catch.

.github/workflows/ci.yml
```diff
@@ -145,3 +147,9 @@ jobs:
           # message to the pull request title
           # https://github.com/cypress-io/github-action/issues/124#issuecomment-653180260
           COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title }}
+
+          # Percy config
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+          PERCY_PARALLEL_TOTAL: 3
```
HonkingGoose, line 153:
```suggestion
```

We could try removing this as well: [^docs]

> `PERCY_PARALLEL_TOTAL`: The total number of parallel builds to expect.
> - In most setups this is automatically discovered and does not need to be set. See Parallel test suites for more info.

[^docs]: https://docs.percy.io/docs/environment-variables#optional

Belco90: Let's try this one, although I'm not sure how Percy can derive this one since it depends on the amount of containers defined in the job.

HonkingGoose (approved):

Belco90:
It looks like it's not working correctly after removing `PERCY_PARALLEL_TOTAL`. I'm double-checking if it was just a coincidence or it actually needs that env var (I believe it's gonna be the latter).

HonkingGoose:
If we need it then we should put it back of course! 😉

Belco90:
> If we need it then we should put it back of course! 😉

It seems it's needed, there are snapshots missing now. I'll revert the parallel env var.

HonkingGoose (approved):

PR #662, opened 2022-01-12
510 add more utils tests

Belco90:
## Changes

- Add tests for all custom utils
- Minor improvements here and there

## Context

Relates to #510, continues #659.

## Tests

I've checked my work by:

- [ ] Testing manually

### DIFF
--- a/src/__tests__/utils.test.ts
+++ b/src/__tests__/utils.test.ts
@@ -1,6 +1,7 @@
-import type { Repository } from '../models'
+import type { ReleaseLike, Repository } from '../models'
 
 import {
+  getReleaseVersion,
   mapRepositoryToQueryParams,
   mapStringToRepositoryQueryParams,
 } from '~/utils'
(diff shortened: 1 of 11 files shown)


src/__tests__/utils.test.ts
```diff
@@ -1,29 +1,290 @@
...
+  `(
+    'should map $label',
+    ({
+      input,
+      output,
+    }: {
+      input: undefined | Repository
+      output: RepositoryQueryParams
+    }) => {
+      const result = mapRepositoryToQueryParams(input)
```
Belco90, line 31: I've moved these tests to `it.each` as spoken in the previous PR.

src/models.ts
```diff
@@ -1,25 +1,21 @@
...
+export type ReleaseGroup = SemVerGroup | MiscGroup | string
 
 export type RepositoryQueryParams = {
   repo: string
   owner: string
 }
 
-export type Repository =
-  RestEndpointMethodTypes['repos']['get']['response']['data']
+export type Repository = components['schemas']['full-repository']
```
Belco90, line 16: I'm using this other type instead since it's more straightforward and easier to find for TS, so I can see the final computed type in my IDE.

src/utils.ts
```diff
@@ -25,29 +22,28 @@ export function mapRepositoryToQueryParams(
...
-  }
-}
-
-type FilterReleasesNodes = {
-  releases: Array<Release>
-  from: ReleaseVersion
-  to: ReleaseVersion
+): RepositoryQueryParams {
+  const [owner = '', repo = ''] = str.split('/')
+  return { owner, repo }
```
Belco90, line 27: I was able to simplify this one after adding tests.

src/__tests__/utils.test.ts
HonkingGoose: The table layout makes the test easier to understand, I can see what the input and outputs are.

I did not know you could use a `label` with that `it.each` test construct. Very nice! 😄

src/__tests__/utils.test.ts
```diff
@@ -1,29 +1,290 @@
...
+
+    expect(result).toEqual([])
   })
 
-  it('should map a full repository', () => {
-    const result = mapRepositoryToQueryParams({
-      owner: { login: 'foo' },
-      name: 'bar',
-    } as Repository)
+  it('should return an empty array if the releases are out of the rage', () => {
```
HonkingGoose, line 126:
```suggestion
  it('should return an empty array if the releases are out of the range', () => {
```

src/utils.ts
HonkingGoose: Having tests allows you to change the code and be sure you're not breaking existing functionality. It looks a lot better now!

HonkingGoose (changes requested):
Found a small typo in one of the descriptions. 😉

HonkingGoose (approved):

PR #679, opened 2022-01-21
Move Visual Testing tool from Percy to Happo

### LINKED ISSUE #511: Visual diff testing

Belco90:
## Changes

- Moving the Visual Testing tool from Percy to Happo

## Context

Closes #511

## Tests

I've checked my work by:

- [ ] Testing manually


Belco90:
I think that Happo isn't compatible with Cypress running in parallel. I'll have to check this again.

Belco90:
> I think that Happo isn't compatible with Cypress running in parallel. I'll have to check this again.

I was wrong. It's doable, I'm updating this PR to make it compatible with parallel builds.

.github/workflows/e2e.yml
```diff
@@ -68,6 +69,9 @@ jobs:
           COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title }}
 
           # Visual Testing config
-          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
-          PERCY_PARALLEL_TOTAL: 3 # remember to keep this synced with matrix.containers
-          PERCY_BRANCH: ${{ github.head_ref }}
+          HAPPO_API_KEY: ${{ secrets.HAPPO_API_KEY }}
+          HAPPO_API_SECRET: ${{ secrets.HAPPO_API_SECRET }}
+
+      - name: Finalize Happo
```
trotzig, line 75: Unrelated to the failure, but this step needs the HAPPO_API_SECRET and HAPPO_API_KEY env variables as well.

.github/workflows/e2e.yml
```diff
@@ -15,8 +15,10 @@ permissions:
 
 jobs:
   cypress:
-    name: Cypress with Percy
+    name: Cypress with Happo
     runs-on: ubuntu-latest
+    env:
+      HAPPO_NONCE: '${{ github.event_name }}-${{ github.sha }}'
```
trotzig, line 21: While confusing, I think this is what's causing the 500 errors. The max length of this string in the database is 50, and I measure this at 53 chars. The fact that we're not validating this on the server is sloppy, I'll make sure we fix that. In the meantime, I think you can change this to 

```yaml
HAPPO_NONCE: '${{ github.run_id }}-${{ github.run_attempt }}'
```

trotzig (approved):
I think the root cause of the failed runs is that HAPPO_NONCE is longer than 50 chars. 

While reviewing, I also found that the `npx happo-e2e finalize` step is run without the HAPPO_API_* env variables. This is a separate issue but needs to be fixed for the whole flow to work. 

Other than that, things look good! Thanks for trying out Happo ❤️

.github/workflows/e2e.yml
trotzig: I've update the API on happo.io to send a 400 error with a useful error message in case the nonce is too long.

.github/workflows/e2e.yml
Belco90: Of course 🤦‍♂️ . They are now shared across the whole job properly. Thanks!

trotzig:
I'm trying to figure out what is causing that new error. 🕵️‍♀️

Belco90:
@trotzig Almost fixed! The 2nd job for Cypress failed: https://github.com/octoclairvoyant/octoclairvoyant-webapp/runs/4927813440?check_suite_focus=true

```
Error: Request to POST https://happo.io/api/async-reports/bd99089b804f2ed2a1daa398596d127de50e08bc/finalize failed: 404 - Async report not found for sha=bd99089b804f2ed2a1daa398596d127de50e08bc, nonce=1741998339-1
    at retries (/home/runner/work/octoclairvoyant-webapp/octoclairvoyant-webapp/node_modules/happo.io/build/makeRequest.js:87:17)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async finalizeAll (/home/runner/work/octoclairvoyant-webapp/octoclairvoyant-webapp/node_modules/happo-e2e/bin/happo-e2e.js:107:3)
    at async init (/home/runner/work/octoclairvoyant-webapp/octoclairvoyant-webapp/node_modules/happo-e2e/bin/happo-e2e.js:197:7) {
  statusCode: 404
}
```

Any idea what could cause that!

trotzig:
The finalize call needs to happen once only, after all parallel test runs have finished. Right now it's run with each individual percy run. I'm looking through github actions docs to see if I can figure out how to configure this.

trotzig:
I think you'll have to use `needs` for this: 
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

I haven't set up parallel Cypress runs with Github Actions. With CircleCI, you use `post_step` which is built into the Cypress ORB I think: https://github.com/happo/happo-cypress/blob/c77ebdd664c4548503558c7e02f497ce90dfc61e/.circleci/config.yml#L83

Belco90:
Ooh I see. Didn't realize the finalize step should happen only once. Indeed, it should be possible in a separate job using `needs` property. I'll give it a try later.

trotzig:
It looks like you solved it -- nice!

Belco90:
@trotzig working now, thank you for your help!

@HonkingGoose I need to setup mobile sizes for the screenshots and this PR will be ready.

.happo.js
```diff
@@ -0,0 +1,18 @@
...
+    'Chrome - Desktop': new RemoteBrowserTarget('chrome', {
+      viewport: '1024x768',
+    }),
+    'Firefox - Desktop': new RemoteBrowserTarget('firefox', {
+      viewport: '1024x768',
+    }),
+    'Chrome - Mobile': new RemoteBrowserTarget('chrome', {
+      viewport: '320x640',
+    }),
+    'Safari - Mobile': new RemoteBrowserTarget('ios-safari'),
```
trotzig, line 16: This will need a `viewport` as well for the build to succeed. You can add any width and height, it will just ignore it. E.g. `viewport: '375x700'`

Belco90: I see, I'll add a viewport here! Thanks.

Belco90:
This is ready to go now!

Belco90:
@trotzig Safari iOS screenshots don't seem to be really accurate. Any idea what could cause that?

trotzig:
I'm having a look at this right now.

trotzig:
I'm assuming you're referring to the start/home page looking odd. Right now Happo runs ios safari 13.5, and this is what the home page looks like in Safari on such a device: 

Screenshot 2022-01-26 at 10 59 15

Compare that with the Happo screenshot: 
Screenshot 2022-01-26 at 11 07 47

It looks like there are two issues: 
a) The image is missing
b) The top container is truncated

I'll investigate a) -- Happo should wait for the image to be ready before taking the screenshot. 

For b) I noticed the container is a reverse column flexbox. There might be bugs around this in iOS Safari <v14. 

It's on our roadmap to upgrade workers to iOS 15, but I can't say for sure when that will happen.

Belco90:
Indeed, I was referring to that page, but the comparator one looks strange too (no padding in the main container for example).

Ooh so it's an older Safari. That makes sense then, it doesn't support flexbox and we are not adding polyfills for it since it's officially not supported.

Thanks for taking a look at this. This is ok then, we will keep an eye on Happo updates to support modern Safari versions.

trotzig:
Apologies, it's iOS Safari 14.3 we're running, not 13.5. So it's old but not ancient. It definitely has flexbox support, but it looks like there was a bug at some point in relation to how the home page container/containers are laid out. 

I decided to dig into this some more and found that the "fix" is quite simple. Just set `flex-basis` to some other value than `0px` on this container: 
Screenshot 2022-01-26 at 11 50 26

This involves configuring chakra stacks and containers in some way. I played around with some different ways of doing it and ended up with a working fix -- just wrap some Stack children in StackItem. Here's a diff you can apply if you want to: 

```
diff --git a/src/pages/index.tsx b/src/pages/index.tsx
index e7d51bd..f5df9eb 100644
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -8,6 +8,7 @@ import {
   ListIcon,
   ListItem,
   Stack,
+  StackItem,
   VStack,
   Text,
   Container,
@@ -36,29 +37,32 @@ const MainSection = () => (
     direction={{ base: 'column-reverse', lg: 'row' }}
   >
     <Stack
-      shouldWrapChildren
       alignItems={{ base: 'center', lg: 'start' }}
       spacing="8"
       flexGrow={1}
       flexBasis={0}
     >
-      <Heading
-        as="h2"
-        color={{ base: 'gray.900' }}
-        fontSize="4xl"
-        fontWeight="black"
-        textAlign={{ base: 'center', lg: 'left' }}
-        letterSpacing="tight"
-      >
-        {BRIEF_DESCRIPTION}
-      </Heading>
-      <Flex py="6" justify="center" align="center">
-        <NextLink href="/comparator" passHref>
-          <Button as="a" variant="cta">
-            Try me now!
-          </Button>
-        </NextLink>
-      </Flex>
+      <StackItem>
+        <Heading
+          as="h2"
+          color={{ base: 'gray.900' }}
+          fontSize="4xl"
+          fontWeight="black"
+          textAlign={{ base: 'center', lg: 'left' }}
+          letterSpacing="tight"
+        >
+          {BRIEF_DESCRIPTION}
+        </Heading>
+      </StackItem>
+      <StackItem>
+        <Flex py="6" justify="center" align="center">
+          <NextLink href="/comparator" passHref>
+            <Button as="a" variant="cta">
+              Try me now!
+            </Button>
+          </NextLink>
+        </Flex>
+      </StackItem>
     </Stack>
     <Image
       priority

```

In the meantime, switch the target to a regular `safari` one with a smaller viewport. That's also running webkit, so there won't be much differences from running in a phone.

HonkingGoose:
Is this a Chrakra UI bug, or something we're doing wrong?

Belco90:
@trotzig oh wow, thanks for the detailed investigation and explanation! Just knowing there was an issue with some unsupported flex property would have been enough, so I really appreciate it.

I'll switch to the regular `safari` then.

trotzig:
@HonkingGoose I'd say it's probably a webkit bug somewhere deep down. One could argue that it's a Chakra bug as well, since using the Stack component gives undesired results in earlier webkit versions.

Belco90:
I've replaced `ios-safari` with `safari`. This should be good to go now 🙏

src/pages/index.tsx
```diff
@@ -40,7 +40,6 @@ const MainSection = () => (
       alignItems={{ base: 'center', lg: 'start' }}
       spacing="8"
       flexGrow={1}
-      flexBasis={0}
```
Belco90, line 43: Just removing this fixes the layout for Safari on iOS 14! Not really sure why I added this property, but it seems like a leftover from a previous layout implementation.

Belco90:
#### 📝 Changed routes:


| [**/**](https://octoclairvoyant-webapp-4dyhieti1-octoclairvoyant-team.vercel.app/) |
|:-:|
|<a href="https://octoclairvoyant-webapp-4dyhieti1-octoclairvoyant-team.vercel.app/">Screenshot of /</a><br /><sup><a href="https://images.weserv.nl?url=https://sjc.microlink.io/ds_Ir-ngTnFve-j_n0p_8nqA26T3Mso3q4UIkKBlWYEnn9uatpDiIL2C8r1GHjJoBXef8DszhXJI8z1jBDOGmw.png">(view full size)</a> |







Commit 1fa8f2127c4a1e863cbed3aff4d302207b4d5b88 (https://octoclairvoyant-webapp-4dyhieti1-octoclairvoyant-team.vercel.app).

cypress/plugins/index.ts
```diff
@@ -9,12 +9,14 @@
...
+// @ts-ignore
+import happoTask from 'happo-cypress/task'
+
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
-
-const pluginConfig: Cypress.PluginConfig = () => {
-  // `on` is used to hook into various events Cypress emits
-  // `config` is the resolved Cypress config
+const pluginConfig: Cypress.PluginConfig = (on) => {
```
HonkingGoose, line 18:
```suggestion
  // `on` is used to hook into various events Cypress emits
const pluginConfig: Cypress.PluginConfig = (on) => {
```

Might be nice to leave the comment in?

Belco90: Sure, let me get the original comment back.

.github/workflows/e2e.yml
```diff
@@ -67,7 +71,26 @@ jobs:
...
           COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title }}
 
-          # Visual Testing config
-          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
-          PERCY_PARALLEL_TOTAL: 3 # remember to keep this synced with matrix.containers
-          PERCY_BRANCH: ${{ github.head_ref }}
+  finalize-happo:
+    name: Finalize Happo
+    runs-on: ubuntu-latest
+    if: ${{ always() }}
```
HonkingGoose, line 77: What does this mean: `if: ${{ always() }}`?

It looks a bit odd, if we always want to run it, couldn't we leave the `if` statement out?

Belco90: TL;DR This is the way GitHub Actions allows you to run a postscript regardless of the previous step result.

Each job in the CI will be run only if the previous job is finished successfully. Some CI tools have a specific property for this, like [`after_script` from Travis CI](https://blog.travis-ci.com/after_script_behavior_changes/).

GitHub Actions doesn't have a specific property to do this, so you need to indicate which is the dependent step with `needs` + indicate to run it no matter what's the result from the previous step with `if: always` to get the same behavior.

We need this behavior to tell Happo to finalize the screenshot process after Cypress tests, no matter what's the result from Cypress itself.

HonkingGoose: > We need this behavior to tell Happo to finalize the screenshot process after Cypress tests, no matter what's the result from Cypress itself.

Ah that's the key bit that I didn't understand from reading the workflow file. 😄

HonkingGoose (approved):

PR #827, opened 2022-04-09
Skip Cypress installation in ci.yml workflow

Belco90:
## Changes

- Sets proper env var to skip Cypress binary installation in `cy.yml` workflow

## Context

There is not ticket for this. I just realized we were installing Cypress binary in every single job for all workflows. This is only necessary in the `e2e.yml` one, so I'm disabling this for `ci.yml` to save time during dependencies installation.

## Tests

I've checked my work by:

- [X] Testing manually

You can see here this is working as expected:
CleanShot 2022-04-09 at 14 37 25


HonkingGoose (approved):

HonkingGoose:
Good idea! 😄 Saves us some time when running the workflow. 🚀

PR #850, opened 2022-04-16
Upgrade framer-motion to v6

Belco90:
## Changes

- Upgrade framer-motion to v6

## Context

We prevented renovate to upgrade framer-motion to v6 since it wasn't compatible with chakra-ui, but it is compatible now.

## Tests

I've checked my work by:

- [X] Testing manually


HonkingGoose (approved):
I think we only use Framer for the spinning fuchsia icon when you're entering the repository name in the comparator screen?

The spinning icon still works, so I think we're good to go. 😄

PR #920, opened 2022-05-16
Upgrade React and Chakra UI

Belco90:
## Changes

- Upgrade React to v18
- Upgrade Chakra UI to v2, replacing deprecated features usages
- Fix nested HTML validation (`div` cannot appear as a descendant of `p`)

## Context

Chakra was the last dependency pending to be compatible with React v18, so I'm upgrading both at the same time since Chakra v2 is only compatible with React v18.

## Tests

I've checked my work by:

- [X] Testing manually


src/pages/index.tsx
```diff
@@ -74,55 +74,55 @@ const FEATURES_DESCRIPTIONS: Array<FeatureItemProps> = [
     icon: HiOutlineSwitchHorizontal,
     title: 'Compare releases easily',
     children: (
-      <Box>
+      <Text as="span">
```
Belco90, line 77: This change isn't fine: it's modifying the UI. I'll address it properly.

Belco90: Should be fixed now

Belco90:
It turns out the Happo screenshots were incorrect before, and after this upgrade they are correct!

Setting the PR as draft tho, since one of the Chakra subdependencies is not 100% compatible with React v18 yet (they'll release it soon).

HonkingGoose:
> It turns out the Happo screenshots were incorrect before, and after this upgrade they are correct!

Yay! 🥳 

> Setting the PR as draft tho, since one of the Chakra subdependencies is not 100% compatible with React v18 yet (they'll release it soon).

In that case it makes sense to wait to have everything compatible with React v18 and be done with it at once. 😄

Belco90:
Chakra UI upgraded to the latest version, which is fully compatible with React v18 🤞

Belco90:
@HonkingGoose this is ready to go!

HonkingGoose (approved):
Looks good to me.

PR #1080, opened 2022-08-12
Indent lines with tabs instead of spaces

Belco90:
## Changes

- Setting prettier to [indent lines with tabs instead of spaces.](https://prettier.io/docs/en/options.html#tabs)
- Formatting all the affected files

## Context

[Prettier is considering switching to tabs by default in v3](https://twitter.com/rich_harris/status/1541761871585464323?s=12&t=Pz1PyMfaUUv8tP2HxYX65w). This is a great change, you can find the motivation behind it here: https://github.com/prettier/prettier/issues/7475

I think we should switch to tabs now to improve the accessibility of the code. The tab size can be customized in GitHub here: https://github.com/settings/appearance

## Tests

I've checked my work by:

- [X] Testing manually


.eslintrc.js
```diff
@@ -2,85 +2,85 @@
...
-    'import/internal-regex': '^(@app-|~)',
-    jest: {
-      version: jestVersion,
-    },
-  },
-  rules: {
-    // Base
-    'no-shadow': 'error',
-    'no-warning-comments': 'off',
+	extends: [
```
HonkingGoose, line 5: I'm okay with this idea. But 4 spaces per tab is too wide for me.

Can you make it so it uses 2 spaces per tab?

HonkingGoose (changes requested):

.eslintrc.js
Belco90: That's one of the advantages of tabs: each user decide the size they want to see for the tabs!

I can't set that, that's up to you in your IDE. As I mentioned in the description, you can customize it for GitHub here: https://github.com/settings/appearance

HonkingGoose: > That's one of the advantages of tabs: each user decide the size they want to see for the tabs!

Oh that's the big idea here!

I forgot to tell GitHub how many spaces it should use for tabs. I've changed my settings on GitHub, and it looks the same as before now. 🥳 

Yeah, let's try this out. In any case it will make merging a Prettier `3.0.0` PR a bit easier, as the files are already formatted in the correct way. 😄

HonkingGoose (approved):

PR #1090, opened 2022-08-16
Enabling TS ESLint advanced configs

Belco90:
## Changes

- Enabling TS ESLint advanced configs
- Fixing new errors reported

## Context

I realized TS ESLint provides some advanced configs: [recommended rules requiring type checking](https://typescript-eslint.io/docs/linting/configs#recommended-requiring-type-checking), and [strict rules](https://typescript-eslint.io/docs/linting/configs#strict).

I've enabled both, disabling some specific rules and fixing new errors reported.

## Tests

I've checked my work by:

- [X] Testing manually

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,16 +2,10 @@
 const jestVersion = require('jest/package.json').version
 
 module.exports = {
-	parserOptions: {
-		tsconfigRootDir: __dirname,
-		project: ['./tsconfig.eslint.json'],
-	},
 	extends: [
 		'eslint:recommended',
 		'next/core-web-vitals',
 		'plugin:@typescript-eslint/recommended',
-		'plugin:@typescript-eslint/recommended-requiring-type-checking',
-		'plugin:@typescript-eslint/strict',
 		'prettier',
 	],
 	plugins: ['unicorn'],
@@ -54,6 +48,19 @@ module.exports = {
 		// TypeScript
 		{
 			files: ['**/*.ts?(x)'],
+			parserOptions: {
+				tsconfigRootDir: __dirname,
+				project: ['./tsconfig.eslint.json'],
+			},
+
+			/*
+			 * Linting with Type Information only for TS files:
+			 * https://typescript-eslint.io/docs/linting/typed-linting/#i-get-errors-telling-me-the-file-must-be-included-in-at-least-one-of-the-projects-provided
+			 */
+			extends: [
+				'plugin:@typescript-eslint/recommended-requiring-type-checking',
+				'plugin:@typescript-eslint/strict',
+			],
 			rules: {
 				'@typescript-eslint/array-type': [
 					'warn',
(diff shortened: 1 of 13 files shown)


.eslintrc.js
```diff
@@ -49,20 +49,36 @@ module.exports = {
...
-					'error',
+					'warn',
 					{
 						default: 'generic',
 					},
 				],
-				'@typescript-eslint/no-floating-promises': 'error',
 				'@typescript-eslint/consistent-type-exports': 'error',
 				'@typescript-eslint/consistent-type-imports': 'error',
-				'@typescript-eslint/no-explicit-any': 'error',
```
Belco90, line 65: These 2 rules removed from here are automatically enabled in the new configs switched on.

additional.d.ts
```diff
@@ -1 +1,10 @@
 // declare here any additional module with no types
+
+declare module 'happo-cypress/task' {
```
Belco90, line 3: I should create a PR in DefinitelyTyped to include this types for Happo. I'll try to do this when I have more spare time.

package.json
```diff
@@ -15,7 +15,7 @@
 		"type-check": "tsc --project tsconfig.validation.json",
 		"type-check:cypress": "tsc --project cypress/tsconfig.json",
 		"lint": "eslint . --ext .js,.jsx,.ts,.tsx --max-warnings 0 --report-unused-disable-directives",
-		"lint:fix": "pnpm lint -- fix",
+		"lint:fix": "pnpm lint --fix",
```
Belco90, line 18: I left a corrupted command in here 🤦 . Fixed now!

src/components/Footer.tsx
```diff
@@ -42,7 +42,7 @@ const Footer = ({ bgColor = 'background2' }: FooterProps) => {
 							aria-label="Powered by Vercel"
 							isExternal
 						>
-							<Image alt="" src={poweredByVercelLogo} />
+							<Image alt="" src={poweredByVercelLogo as string} />
```
Belco90, line 45: SVG files imported within a Next.js project are typed as `any`. I can't override that unless I declare a custom `next-env.d.ts` overriding the types for `*.svg` files. I don't like this idea, since we could miss other type definitions added in the future. Furthermore, I'm casting it to string manually when needed.

More info about this problem here: https://duncanleung.com/next-js-typescript-svg-any-module-declaration/

additional.d.ts
HonkingGoose: That's okay! I'll let you do that in your own time. 😉

src/components/Footer.tsx
HonkingGoose: Let's keep things simple and just cast it to `string` manually when needed. We don't have that many SVGs anyways. 😄

HonkingGoose (approved):
I don't know enough about this to properly review your work. So I'm going to trust that this is okay. 😉

PR #1097, opened 2022-08-17
Improve loading workflow when processing the results

### LINKED ISSUE #531: Move comparator context to external global state

We need to improve the app performance by moving the repo info + filters applied from React Contexto to a proper global state solution.

The state manager chosen would be [zustand](https://github.com/pmndrs/zustand).

Right now, every change to the repo selected or filters applied is causing re-renders everywhere. This affects the app performance. You can notice that when selecting a huge range of versions to compare: for a bit you can see an error message that blinks right before showing the results.

Yes

_No response_

### LINKED ISSUE #859: Show loading spinner when fetching releases

Show a loading spinner in the main section of the comparator while fetching the releases. Right now, there is no specific element to let the user know that releases are being fetched. A good example for this is #831 

Make more explicit to users that releases are still being fetched.

Yes

_No response_

Belco90:
## Changes

- Show loading spinner when fetching releases
- Avoid showing blink error alert before rendering results

## Context

Fixes #531
Fixes #859

## Tests

I've checked my work by:

- [X] Testing manually (accessing to the review environment with `/comparator?repo=renovatebot%2Frenovate&from=32.52.2&to=latest` appended to the URL)


src/components/ComparatorChangelogResults.tsx
```diff
@@ -0,0 +1,119 @@
...
+		</Box>
+	)
+}
+
+interface ComparatorChangelogResultsProps {
+	releases: Array<Release> | null
+	repository: Repository
+}
+
+const ComparatorChangelogResults = ({
```
Belco90, line 66: This component has been extracted from `RepositoryReleasesChangelog.tsx`. With this improvement, I have more control over the workflow of rendering results, errors or loading states.

src/hooks/useProcessReleases.ts
```diff
@@ -115,22 +104,21 @@ function useProcessReleases(
 ): UseProcessReleasesReturn {
 	const [processedReleases, setProcessedReleases] =
 		useState<ProcessedReleasesCollection | null>(null)
-	const [isProcessing, setIsProcessing] = useState(false)
+	const [isProcessing, setIsProcessing] = useState(true)
```
Belco90, line 107: This was another problem: the initial state must be true, so it starts as loading. That was the actual cause of the error blink in the results page.

I've reworked the hook so it's easier to follow.

src/components/RepositoryReleasesChangelog.tsx
```diff
@@ -1,163 +1,72 @@
...
-				</>
+			{(isFetching || isFilteringReleases) && (
+				<Flex align="center" justify="center" height="100%">
+					<CircularProgress
+						isIndeterminate
+						size="8"
+						color="primary.400"
+						aria-label="Loading releases"
+					/>
+				</Flex>
```
HonkingGoose, line 60: I think this is the part that controls the "spinner" in the changelog part of the page?

Can you add a comment like: "Changelog spinner" so that we can find it quickly later?

HonkingGoose (commented):
I'm impressed! The loading state looks way way better now! 🥳

src/components/RepositoryReleasesChangelog.tsx
Belco90: Yes, this is the one. I can do that for sure!

I was actually thinking about showing a loading skeleton instead of the spinner. Let me change it so we can decide which option looks better.

Belco90: Change pushed, you can check the skeleton in the review environment.

Here you have both versions:

**With spinner**

https://user-images.githubusercontent.com/2677072/185387820-977d4b16-bb6a-4b5b-b5af-c183e5dec1e8.mp4

**With skeleton**

https://user-images.githubusercontent.com/2677072/185387851-6dfc7f43-e954-4f69-b55e-7939ebb04777.mp4

---

I prefer the skeleton since it helps guess what sort of content is coming, and it gives the feeling is loading faster.

Belco90: I've pushed another change. I'm actually using both the spinner and the skeleton now:

- if you select the repository through the UI, then the spinner will appear while fetching the releases, and disappear when it's ready for the user to pick the desired versions.
- if you land in the comparator with a preloaded repo + versions from the URL query, then the skeleton will appear while fetching and processing the releases, and disappear when it's ready to display the final changelog (no user interaction needed)

Let me know what you think!

HonkingGoose: I like that we use the spinner _and_ the skeleton now. 😄 

I'm still seeing the flash of the red "processing releases" warning though, when I try it on the Vercel preview:

`/comparator?repo=renovatebot%2Frenovate&from=32.52.2&to=latest`

Belco90: Let me check that!

Belco90: @HonkingGoose It should be fine now!

HonkingGoose (approved):

Issue #741, opened 2022-02-16
Increase and enforce GitHub API release fetching quotas


HonkingGoose:
### Describe the bug

I don't see a changelog after clicking on a legacy style URL to Octoclairvoyant.

### Give the steps to reproduce

1. Click on this link https://octoclairvoyant.vercel.app/comparator?repo=renovatebot%2Frenovate&to=24.22.3&from=23.95.0
2. No changelog

### What browsers are you seeing the problem on?

Bug is not about browser

### Have you thought of a possible solution?

No

### Do you want to help fix the bug?

No

### Is there anything else we need to know?

Maybe the version range is too big, and Octoclairvoyant craches or exceeds some GitHub API limit?

Or old style link not properly converted to new comparator slug?

I got the link from @Belco90 themselves, so I assume the link was good at the time it was posted. 😉 

Source of link: https://github.com/renovatebot/renovate/issues/8177#issuecomment-762352681

Belco90:
This is an interesting bug, let me explain what's happening.

Since most of the repos I tend to check when I built the web app had just a bunch of releases, I'm doing max of 10 requests, with 100 releases per request, which gives us a total of 1,000 releases. You can see the code for this behavior here: https://github.com/octoclairvoyant/octoclairvoyant-webapp/blob/main/src/queries/release.tsx

However, Renobate bot has thousands of releases! So basically what's happening is that the desired From release is out of the first 1,000 releases at this point. You can see this by scrolling to the very bottom of the releases select:
![CleanShot 2022-02-16 at 13 59 04](https://user-images.githubusercontent.com/2677072/154269650-5d58ed0d-77e4-44ca-b2a4-adcb731a07ff.png)

So we need to update the login within that query so it keeps asking for more releases if after those 10 iterations both of them are not available yet.

HonkingGoose:
Thanks you for the explanation. So the problem is with the GitHub requests limit, not with the URL. 😄 

What if users spam our site with queries like from `v1.0.0` to `31.84.1` in order to fetch 5000+ releases, plus heavy computation load to cobble it all together? At some point we should ask our users to use their own GitHub API quota instead of ours. 😉 

We should test that:

- The bug is fixed
- We enforce reasonable API limits for "free queries"
- Show modal once user reaches limits, or does one very large query

What do you think?

Belco90:
I'm working on fixing this issue, and I discovered something else:

[...]

PR #1121, opened 2022-08-23
Stubbing Cypress E2E requests to GitHub

### LINKED ISSUE #443: Cypress E2E: stub request per test basics with hardcoded values

It's probably a good idea to stub the request/responses from GitHub.

This a todo left over from PR #378.

Speed up tests, and reduces API requests footprint of running the tests locally.

Yes

I want to help, but don't know how. 🙈 
For now I'll rely on @Belco90 to add this.

Belco90:
## Changes

- Stubbing request for most of the Cypress E2E, leaving the happy path connected to the actual GitHub API
- Fixing an error causing unnecessary requests

## Context

Closes #443 

## Tests

I've checked my work by:

- [X] Testing manually


src/queries/release.tsx
```diff
@@ -46,7 +46,7 @@ function useReleasesQuery(
 				releases.push(...response.data.filter(isStableRelease))
 
 				const isMaxAutoPaginationReached =
-					paginationCount === MAX_AUTO_PAGINATION
+					paginationCount >= MAX_AUTO_PAGINATION
```
Belco90, line 49: This stupid mistake was making the comparator to request way more pages than necessary when needing more than first 1000 results.

The ability to keep paginating was introduced in #741. However, it was suspiciously slow for the given example `/comparator?repo=renovatebot%2Frenovate&to=24.22.3&from=23.95.0`

If you try that preloaded URL in the review environment for this PR, you'll notice the changelog comparison is loaded much faster than before (e.g. production right now).

@HonkingGoose just FYI: when this gets merged, renovate comparison should be much faster.

HonkingGoose: Good catch! 🥳

PR #1104, opened 2022-08-19
Group development dependencies

HonkingGoose:
## Changes

- Use the `npm:unpublishSafe` preset instead of having our own custom `packageRule`
- Group all development dependencies, and update them once a month
- Order the `packageRules` by importance

## Context

The `packageRules` config option docs say:

> Important to know: Renovate will evaluate all `packageRules` and not stop once it gets a first match. You should order your `packageRules` in order of importance so that later rules can override settings from earlier rules if needed.

Closes #673.

## Tests

I've checked my work by:

- [ ] Adding new tests or adjusting existing tests
- [ ] Testing manually
- [x] Code review

PR #1204, opened 2022-10-07
Use `npm:unpublishSafe` preset

HonkingGoose:
## Changes

- Use `npm:unpublishSafe` preset [^docs]

## Context

I see you've reverted #1104, which also included this change. I think we should use the preset instead of making our own rule. 😉 

## Tests

I've checked my work by:

- [ ] Testing manually
- [x] Code inspection only

[^docs]: https://docs.renovatebot.com/presets-npm/#npmunpublishsafe

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,6 +3,7 @@
 		"config:base",
 		"helpers:pinGitHubActionDigests",
 		":automergeTypes"
+		"npm:unpublishSafe"
 	],
 	"prCreation": "not-pending",
 	"lockFileMaintenance": {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -3,6 +3,7 @@
 		"config:base",
 		"helpers:pinGitHubActionDigests",
 		":automergeTypes"
```
HonkingGoose, line 5:
```suggestion
		":automergeTypes",
```

Belco90 (approved):
I didn't realize about this! I tried to look for the docs about this preset but can't find anything.

PR #1221, opened 2022-10-11
Cache pnpm dependencies properly

Belco90:
## Changes

- Adds a prior job in CI to install dependencies

## Context

Since all jobs are installing dependencies with pnpm in parallel, they are not able to reuse the cache since only one of them can save it. [See this CI run as an example](https://github.com/octoclairvoyant/octoclairvoyant-webapp/actions/runs/3224811427): the "format check" job was able to save the pnpm cache in "post setup node" step, but none of them were able to restore any cache.

We are going to try mitigating this by adding a prior job which installs the dependencies, then all CI jobs depend on it to take advantage of the generated cache.

## Tests

I've checked my work by:

- [X] Testing manually

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,27 @@ env:
   CYPRESS_INSTALL_BINARY: 0
 
 jobs:
+  # Putting a first "install" job first, so it saves the pnpm cache for other jobs.
+  install:
+    name: Install dependencies
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


Belco90:
This doesn't bring any noticeable improvement. The CI run mentioned in the ticket description wasn't cached properly because it modified the lock file by updating dependencies. When CI runs without modifying the lock file, the cache works as expected ([like in this CI run](https://github.com/octoclairvoyant/octoclairvoyant-webapp/actions/runs/3113625025)).

PR #1329, opened 2022-11-19
Disable useFullPageFallbackForTallScreenshots in Happo

trotzig:
This commit fixes an issue with timeouts coming from Happo when taking screenshots in Safari.

For tall screenshots (over 4000px), Happo will apply a workaround when taking the screenshot to work around quirks that automated browsers apply. It involves taking a full page screenshot and then cropping the screenshot down to the right size. However, in many cases, the cropping will never happen (when you actually want all the content on the page).

In Safari, we've seen this process take longer than 30s. In these cases, a timeout error is thrown which makes the job eventually fail.

More information about the configuration option here: https://docs.happo.io/docs/configuration#target-usefullpagefallbackfortallscreenshots

By setting useFullPageFallbackForTallScreenshots to false, we can disable the full page+cropping behavior. I've verified that this change makes screenshots finish in less than a few seconds. And the screenshot results look the same.

It's likely that this commit will trigger Happo diffs on some screenshots, mostly because the process to generate a hash from the image content is different in the two codepaths.

## Changes

-

## Tests

I've checked my work by:

- [x] Testing manually

### DIFF
--- a/.happo.js
+++ b/.happo.js
@@ -9,15 +9,19 @@ module.exports = {
 	targets: {
 		'Chrome - Desktop': new RemoteBrowserTarget('chrome', {
 			viewport: DESKTOP_VIEWPORT,
+			useFullPageFallbackForTallScreenshots: false,
 		}),
 		'Firefox - Desktop': new RemoteBrowserTarget('firefox', {
 			viewport: DESKTOP_VIEWPORT,
+			useFullPageFallbackForTallScreenshots: false,
 		}),
 		'Chrome - Mobile': new RemoteBrowserTarget('chrome', {
 			viewport: MOBILE_VIEWPORT,
+			useFullPageFallbackForTallScreenshots: false,
 		}),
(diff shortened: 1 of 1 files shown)


Belco90:
Great, thanks! It makes sense, didn't realized about this option.

However, I see Happo job failing with error "waiting for status to be reported". Any clue about that @trotzig?

trotzig:
I think it's because the HAPPO_API_KEY and HAPPO_API_SECRET aren't made available to CI runs for outside contributors. If you want to you could probably cherry-pick my commit into a local branch that you then push as a PR. That way the secrets will be there when the CI job runs. Or just reproduce the changes I made in a commit of your own (I don't care about attribution here 🙂).

Belco90:
Aah I see! I'll just merge it then, and check the behaviour in other PRs.

Belco90 (approved):

Belco90:
It's working fine, many thanks!

PR #1665, opened 2023-04-03
Migrate to pnpm v8

Belco90:
## Changes

- Update pnpm to v8 in corepack
- Remove default settings
- Regenate lock

## Context

n/a

## Tests

I've checked my work by:

- [X] Testing manually

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,2 @@
 save-prefix=''
-auto-install-peers=true
+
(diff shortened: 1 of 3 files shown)


.npmrc
```diff
@@ -1,2 +1,2 @@
 save-prefix=''
-auto-install-peers=true
```
Belco90, line 2: This is enabled by default in v8.

PR #2198, opened 2024-01-12
Set up CodeQL to v3 using default setup

Belco90:
## Changes

- Removes custom setup using v2
- Set ups CodeQL using default config, which is automatically updated to v3

## Context

[GitHub announced the deprecation of CodeQL Action v2](https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/). Users of code scanning default setup don't have to do anything since it automatically uses v3.

Since we weren't taking advantage of the advanced setup, this PR removes that config in favor of the default one that gets updated and configured automatically.

CleanShot 2024-01-12 at 17 19 07

## Tests

I've checked my work by:

- [X] Testing manually


Belco90:
[This is a pipeline run with the CodeQL v3 after using the default config](https://github.com/octoclairvoyant/octoclairvoyant-webapp/actions/runs/7504711945/job/20432386484?pr=2198).

